PR #4, opened 2022-05-09
New: Introduce Section field type and replace current use of Groups

jonwaldstein:
## Description

The Fields API term `Group` was meant to be defined as an abstract.  Currently it's being used to contain sections of fields. 
 To closer match our terms & expectations of the form builder, this PR introduces a new `Section` type that extends `Group`.  This will be how we define separate field sections moving forward.  Eventually this will also be moved into core (hence the same namespace).

The api for using `Section` is the same as `Group`.

```
Section::make('donationDetails')
  ->label(__('Donation Details', 'give'))
  ->append(
      Text::make('amount')
          ->label(__('Donation Amount', 'give'))
          ->defaultValue(50)
          ->required()
  )
  ```

## Affects

The current use of the fields api & how the form block is being rendered.

## Testing Instructions

- Pull the branch, npm run dev, and make sure the form looks the same.

## Pre-review Checklist

-   [x] Includes unit and/or end-to-end tests

### DIFF
--- a/src/Framework/FieldsAPI/Section.php
+++ b/src/Framework/FieldsAPI/Section.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Give\Framework\FieldsAPI;
+
+/**
+ * @unreleased
+ */
+class Section extends Group
+{
+    /**
+     * @unreleased
+     */
+    const TYPE = 'section';
+}
--- a/src/Framework/FieldsAPI/Section.php
+++ b/src/Framework/FieldsAPI/Section.php
@@ -11,17 +11,4 @@ class Section extends Group
      * @unreleased
      */
     const TYPE = 'section';
-
-    /**
-     * @unreleased
-     *
-     * @param  Field  ...$fields
-     * @return $this
-     */
-    public function addFields(Field ...$fields): Section
-    {
-        $this->append(...$fields);
-
-        return $this;
-    }
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jonwaldstein** [comment]

cc: @JasonTheAdams

File: src/Framework/FieldsAPI/Section.php [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+     */
+    const TYPE = 'section';
+
+    /**
+     * @unreleased
+     *
+     * @param  Field  ...$fields
+     * @return $this
+     */
+    public function addFields(Field ...$fields): Section
```
JasonTheAdams, line 21: Probably should just be a node as it can be a `Field`, `Element`, or `Group`. I think the `Group` parent class already has a method that appends nodes.

JasonTheAdams (changes requested):
Pretty much there. Just one note.

src/Framework/FieldsAPI/Section.php
jonwaldstein: Yes `Group` already has `append`.  I thought we talked about making it more specific with `Section` like I was doing here but If that's not the case I can just remove this and use append.

JasonTheAdams: I don't really see the value right now. Let's add something like this later if a use-case comes up.

jonwaldstein: Okay, just reverted 6eacbd97

PR #5, opened 2022-05-09
Feature: Form Builder prototype

kjohnson:
## Description

This PR introduces the Form Builder (aka Givenberg) as a mono-repo package.

## Testing Instructions

From the root directory of the plugin, the `@givewp/form-builder` can be started using the following command, and can be accessed at `http://localhost:3000/` for the development build.

```
npm run start --workspace=@givewp/form-builder
```

Accessing the form builder from within the plugin requires that the `@givewp/form-builder` package first be built with the following command, and can then be access from the WordPress admin under `Donations -> Form Builder`:

```
npm run build --workspace=@givewp/form-builder
```

## Visuals

image

## Pre-review Checklist


packages/form-builder/package.json
```diff
@@ -0,0 +1,152 @@
+{
+  "name": "@givewp/form-builder",
+  "private": true,
+  "homepage": ".",
+  "scripts": {
+    "start": "node scripts/start.js",
+    "build": "node scripts/build.js",
+    "test": "node scripts/test.js"
```
jonwaldstein, line 8: @kjohnson i'm a little concerned with the amount of bundle config setup here. Mainly for maintenance reasons, this would be very hard to keep up with. Is there any reason we can't just use something like Laravel mix or even vite?

A note on workspaces, I think it's a fine place to start with this but we may want to consider turborepo in the future.

jonwaldstein (approved):
@kjohnson this is a great start! I can tell you really went into the depths of Gutenberg to make this functional so kudos 🙌 .  I'm going to approve this so we can get this prototype into develop and move things along.  My only real concern at this point is the build process which can be improved down the line as well as many other pieces that we already have plans for.

kjohnson:
@jonwaldstein the build process is just Create React App. Happy to re-tool later if that is what we want to do.

PR #9, opened 2022-05-27
New: introduce form registrars and shared types

JasonTheAdams:
## Description

The goal of this PR is to set up all the necessary registrars in order to get a Next Gen form up and running. At this time, there are two registrars:
1. Gateways
2. Template

The Gateways registrar already existed, so it's been refactored to exist in a shared space. The Template registrar exists, but only just, and will be built further out in a subsequent PR.

There are also a lot of shared TS Interfaces throughout, so the common ones have been moved to a shared location. These can be imported from `@givewp/forms/types`.

Finally, quite a bit of refactoring was introduced to the gateways, both to implement the new registrar structure, but also to take advantage of TypeScript in the gateways themselves.

## Affects

How Next Gen forms are loaded, including gateways

## Testing Instructions

This is mostly a code review, but make sure the form still works as far along as it did.

## Pre-review Checklist


JasonTheAdams:
@jonwaldstein This is ready for review! Let me know if you have any questions!

src/NextGen/Gateways/NextGenTestGateway/nextGenTestGateway.tsx
```diff
@@ -1,19 +1,23 @@
 import {__} from '@wordpress/i18n';
+import type {FormData, Gateway} from '@givewp/forms/types';
 
-window.givewp.gateways.register({
+const gateway: Gateway = {
     id: 'test-gateway-next-gen',
-    beforeCreatePayment(values) {
+    label: 'Test Gateway',
```
jonwaldstein, line 6: As far as the Gateway _type_ is concerned, we want it to be aware that it has a label.  Although, the gateway label needs to be set from the server (as we're doing in the gateway registrar) because it can be customized in the backend. So, we don't want gateways to think they can hard code a label here since it won't actually do anything.  What do you think we should do?

JasonTheAdams: Resolved per our discussion in 4a5b7ea

jonwaldstein (approved):
@JasonTheAdams i'm really glad you read the typescript docs because this is awesome. 🤩

PR #11, opened 2022-06-03
New: Next Gen Stripe Gateway

jonwaldstein:
## Description

To really put the Next Gen gateway api to the test, we need a tangible gateway.  Since Stripe is usually the most involved in terms of front-end, we chose to scaffold a new Stripe gateway using the [Payment Element api](https://stripe.com/docs/payments/payment-element).  The PR adds the new gateway and will provide enough functionality to complete a donation on the next gen form.  There is still more polish to do in regards to the full transaction cycle, including receipts - but those pieces will come soon after this.

**JS Gateway Updates**
- New method `afterCreatePayment?(response: object): Promise<void> | Error` that can be used after creating the donation & returning from `createPayment($donation)` from the gateway.

## Affects

- The js gateway api
- Replace axios with fetch

## Visuals

Screen Shot 2022-06-03 at 2 51 17 PM

## Testing Instructions

- Run `npm run dev`
- Make sure your stripe account is setup & next gen stripe enabled
- Donate using the stripe gateway using the next gen donation form block

## Pre-review Checklist


jonwaldstein:
@JasonTheAdams please review this one.  There is still more polish to do in regards to the full transaction cycle (including receipts) & the api in general - but those pieces will come soon after this.

src/NextGen/DonationForm/Blocks/DonationFormBlock/app/utilities/postData.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @unreleased
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options
+ */
+export default async function postData(url: string, data: object = {})  {
+  // Default options are marked with *
+  const response = await fetch(url, {
+    method: 'POST', // *GET, POST, PUT, DELETE, etc.
+    mode: 'cors', // no-cors, *cors, same-origin
```
JasonTheAdams, line 9: Don't many/most of the requests go to the same origin? I'm not sure we really need to hard-code this.

src/NextGen/DonationForm/resources/types.ts
```diff
@@ -75,6 +75,11 @@ export interface Gateway {
      * A hook before the form is submitted.
      */
     beforeCreatePayment?(values: FormData): Promise<object> | Error;
+
+    /**
+     * A hook after the form is submitted.
+     */
+    afterCreatePayment?(response: object): Promise<void> | Error;
```
JasonTheAdams, line 82: Is there a more specific name we can give these? What are the gateways intended to do at these points? Perhaps this is fine. It's just not clear to me.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @throws InvalidPropertyName
+     */
+    protected function getPaymentIntentArgsFromDonation(Donation $donation, Customer $customer)
+   {
+       // Collect intent args to be updated
+        $intentArgs = [
+            'amount' => $donation->amount->formatToMinorAmount(),
+            'customer' => $customer->id,
+            'description' => (Call::invoke(SaveDonationSummary::class, $donation))->getSummaryWithDonor(),
```
JasonTheAdams, line 100: Now that we're on PHP 7.0, I don't think we actually need to use `Call::invoke` anymore. It will also track itself better if we don't in the IDE.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
```diff
@@ -0,0 +1,147 @@
...
+        DonationNote::create([
+            'donationId' => $donation->id,
+            'content' => sprintf(__('Stripe Payment Intent Client Secret: %s', 'give'), $intent->client_secret)
+        ]);
+
+        give_update_meta(
+            $donation->id,
+            '_give_stripe_payment_intent_client_secret',
+            $intent->client_secret
+        );
```
JasonTheAdams, line 145: Mmm, good reminder for us to finish up the meta stuff in the models.

JasonTheAdams (changes requested):
Lovely! Did a quick code review. I'll test this after the comments are addressed. 👍

src/NextGen/DonationForm/Blocks/DonationFormBlock/app/utilities/postData.ts
jonwaldstein: Yeah, I think same-origin is fine.  `mode: 'cors'` is actually the default (noted by * in comments) so we would need to mark this as `same-origin`.

src/NextGen/DonationForm/resources/types.ts
jonwaldstein: I thought this was a specific name lol.  `beforeCreatePayment()` fires before the gateway class method `createPayment()`.  `afterCreatePayment()` gets fired after  the gateway class method `createPayment()`.  This is where the gateway could return data to itself using `ReturnToBrowser` / json and do something with that response.  

In the case for Stripe it needs to return a stripe intent status to the front-end and do something with it.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
jonwaldstein: yes your right! c53b0225

src/NextGen/DonationForm/Blocks/DonationFormBlock/app/utilities/postData.ts
jonwaldstein: Changed mode to `same-origin` 0f13c2e4 8e305df3

Let's keep this config with the comments in tact for now. I want to make sure it's setup appropriately as we introduce more gateways.

jonwaldstein:
@JasonTheAdams comments addressed, give her a whirl 🧪

src/NextGen/DonationForm/resources/types.ts
JasonTheAdams: Perhaps it is a specific as it can be. Right now the naming caters towards the event — that is, the name is clear that it's happening at a point in time and when. That doesn't really inform the gateway as what it ought to be doing at that time. I don't know if this is actually accurate, but something like, `prepareForDonation` and `finishDonation` — the point being that it gently guides the gateway as what it ought to be doing.

That said, perhaps what's being done at this event varies too much from gateway to gateway, in which case the current event-based naming is fine.

jonwaldstein: Gotchya, I think I gravitated toward the event-based naming because it coincides with the backend api naming convention and was easier to conceptualize in what order they get triggered while writing the logic.  I know what you're saying about guiding the intention of the methods but as you said, the intention could vary between gateways.  Some Gateways might need even need to use these.

Some use cased I see for each:

**beforeCreatePayment()**
- sending data to `createPayment()`, validation, gateway object mounting, pre-fetching something/using custom route

**afterCreatePayment()**
- receiving data from `createPayment()`,  gateway confirmation, redirecting (the framework will probably have a general way of redirecting to receipt as well).

JasonTheAdams (approved):
It makes a donation with the new Stripe Elements and takes me to the confirmation page. Noice. 💯

PR #32, opened 2022-07-12
refactor: Update header components

kjohnson:
Resolves #

## Pre-review Checklist

-   [ ] Includes unit tests


packages/form-builder/src/App.js
```diff
@@ -70,7 +70,9 @@ function App() {
                         <InterfaceSkeleton
                             header={<Header
                                 saveCallback={saveCallback}
+                                showSecondarySidebar={showSecondarySidebar}
                                 toggleSecondarySidebar={toggleSecondarySidebar}
+                                showSidebar={showSidebar}
```
kjohnson, line 75: @jonwaldstein This will probably need to be refactored to a context hook for builder state. I'm looking into that for a later change.Thoughts?

kjohnson:
@jonwaldstein [flex-basis](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis) looks cool - hadn't seen that before.

packages/form-builder/src/App.js
jonwaldstein: Yeah i'm thinking there will be an application provider for managing this kind of stuff, what you have is fine for now though.

jonwaldstein (approved):
@kjohnson I made a few updates here:

1. Header css updates to make the 3 parts more consistent widths.
2. Separated the Header component into presentational and container components.  I like to do this as much as possible to isolate components that can truly be pure components, where others are functional.
3. Some folder structure updates, moved some things out of the "components" directory, created a "common" directory, and a few root entry points so we can get components & common things easier.  

We're getting to a point where the application is growing and it's going to be really important to keep things organized the best we can.  Speaking from experience, these react apps can take a shape of their own if we're not careful 😁.

Otherwise, great job!

PR #33, opened 2022-07-12
Feature: Form template registration

jonwaldstein:
## Description

We recently [introduced](https://github.com/impress-org/givewp-next-gen/pull/18) the initial template api as it relates to customizing the front-end.  This PR accompanies that by adding a template registration system so new Templates can get registered and pull in their front-end assets. 

One of the goals for Templates, is keeping it as simple as possible.  If someone wants to create a template that just pulls in some CSS we want them to get up and running easily.  Here is a birds eye view of a template simplicity using the (_work-in-progress_) classic template as an example:

```
class ClassicFormTemplate extends FormTemplate {

    public static function id(): string
    {
        return 'classic';
    }

    public static function name(): string
    {
        return __('Classic Template', 'give');
    }

    public function css(): string
    {
        return GIVE_NEXT_GEN_URL . 'src/NextGen/DonationForm/FormTemplates/ClassicFormTemplate/template.css';
    }

    public function js(): string
    {
        return GIVE_NEXT_GEN_URL . 'src/NextGen/DonationForm/FormTemplates/ClassicFormTemplate/template.js';
    }
}

add_action('givewp_register_form_template', function (FormTemplateRegistrar $formTemplateRegistrar) {
    $formTemplateRegistrar->registerTemplate(ClassicFormTemplate::class);
});
```

## Affects

The new "classic form" was created as the new default for next gen forms.  Viewing the form should reflect some basic CSS to show the template is active.  Note - this PR is not intended to have a complete version of the _Classic Template_ - just using it for a tangible example until we're ready to build it out.

Also added phpunit here out of necessity.

## Visuals

Screen Shot 2022-07-13 at 5 37 43 PM

## Testing Instructions

View a NextGen form and see some basic css applied.  This is the start of what will soon be the _Classic Template_.

## Pre-review Checklist

-   [x] Includes unit tests


src/NextGen/DonationForm/Blocks/DonationFormBlock/Block.php
```diff
@@ -242,6 +236,30 @@ private function enqueueScripts(int $formId)
...
+            if ($template->css()) {
+                wp_enqueue_style('givewp-form-template-' . $template->getId(), $template->css());
+            }
+
+            if ($template->js()) {
+                wp_enqueue_script(
+                    'givewp-form-template-' . $template->getId(),
+                    $template->js(),
+                    ['give-donation-form-registrars-js'],
+                    false,
```
kjohnson, line 256: This does seem to prevent defining dependencies for the script, which may or may not be needed.

src/NextGen/DonationForm/Blocks/DonationFormBlock/Block.php
```diff
@@ -242,6 +236,30 @@ private function enqueueScripts(int $formId)
...
+        // silently fail if template is missing for some reason
+        if ($formTemplateRegistrar->hasTemplate($formTemplateId)) {
+            $template = $formTemplateRegistrar->getTemplate($formTemplateId);
+
+            if ($template->css()) {
+                wp_enqueue_style('givewp-form-template-' . $template->getId(), $template->css());
+            }
+
+            if ($template->js()) {
+                wp_enqueue_script(
```
kjohnson, line 252: Is there a particular reason for using `wp_enqueue_script` here, as opposed to the `EnqueueScript` classed used elsewhere in this file?

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @unreleased
+     *
+     * @throws InvalidArgumentException
+     */
+    public function getTemplate(string $id): FormTemplate
+    {
+        if (!$this->hasTemplate($id)) {
```
kjohnson, line 32: This check is redundent, as `give()` will throw this exception itself.

```
Give\Framework\Exceptions\Primitives\InvalidArgumentException with message 'Target class [Foo] does not exist.'
```

Instead, we could catch it and re-throw a more specific exception, like Form Template class does not exist.

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @unreleased
+     */
+    public function registerTemplate(string $templateClass)
+    {
+        try {
+            $this->register($templateClass);
+        } catch (InvalidArgumentException $invalidArgumentException) {
```
kjohnson, line 57: What is the reason for catching these two exceptions, specifically, and logging them the same? Should we not just catch a more generic exception?

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+namespace Give\NextGen\Framework\FormTemplates\Registrars;
+
+use Give\Framework\Exceptions\Primitives\InvalidArgumentException;
+use Give\Framework\PaymentGateways\Exceptions\OverflowException;
```
kjohnson, line 6: Should we be re-using Payment Gateway exceptions here?

src/NextGen/Framework/FormTemplates/ServiceProvider.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function boot()
+    {
+        $formTemplateRegistrar = give(FormTemplateRegistrar::class);
+
+        Hooks::doAction('givewp_register_form_template', $formTemplateRegistrar);
```
kjohnson, line 30: Side note: Maybe we should consider overloading the signature for the `Hooks` class methods, so we could pass multiple actions and avoid intermediary variables. Just a thought, but that would be in core.

kjohnson (commented):
@jonwaldstein added my review - mostly questions and additional commentary. Otherwise, not opposed to this getting merged soon, so I submitted the review as a "Comment" as opposed to "Request Changes".

src/NextGen/DonationForm/Blocks/DonationFormBlock/Block.php
jonwaldstein: The `EnqueueScript` class requires defining urls like `$pluginDirPath` and `$pluginDirUrl` which would change depending on if the template is coming from a custom plugin.  The PaymentGateway class is enforcing using that but we want to keep the template simple for people by just specifying a path.

src/NextGen/DonationForm/FormTemplates/ClassicFormTemplate/template.js
```diff
@@ -0,0 +1,3 @@
+window.givewp.template.mount({
+    id: 'classic'
```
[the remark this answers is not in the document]
jonwaldstein: This is only necessary if you want to use the js template api.

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
jonwaldstein: This is just checking if the ID/key exists within the registrar, not if the class exists.

kjohnson: Ah, fair point.

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
jonwaldstein: nope lol updated with new one (but we should probably add OverflowException to our give exceptions)

https://github.com/impress-org/givewp-next-gen/compare/a5e96c062bf7...73b20128278e

src/NextGen/Framework/FormTemplates/Registrars/FormTemplateRegistrar.php
jonwaldstein: Oh yeah, so I was experimenting with _silently_ failing (not actually throwing an error but logging it) here instead of our usual fatal exceptions.  I've been thinking about this lately for our public api's.  From an application standpoint it makes total sense to throw an exception when warranted. On the other hand, we are a distributed application which makes it potentially fatal to a customer's website (we've seen plenty of exceptions and tiny signature changes break our core functionality).  

For templates specifically, that could mean preventing a donation form from loading - which begs the question.. if there's a template conflict will a log suffice or do we really need to _throw_ an exception?

This is perhaps a longer discussion that id like to open up to the team for thoughts.

kjohnson: Let's have that discussion when @JasonTheAdams gets back - I think he'll want input - but go ahead and merge this PR as an incremental step.

PR #39, opened 2022-07-19
Form Builder now supports adding fields via drag-and-drop

### LINKED ISSUE #37: Add drag-and-drop block list sidebar
## Issue Description

Admins should have the ability to click the "+" icon in the top header, see the available block list, and drag them into the appropriate sections.

Image

- [x] Block list should show registered blocks in the appropriate sections
- [x] Admins should only be allowed to drag blocks into sections

Note: searching is _not_ required by the Town Hall, but if it's not a huge complexity to add it in this PR then go ahead and add it.

kjohnson:
Resolves #37

## Description

This PR recreates the drag-and-drop sidebar of the Block Editor, with the notable difference of supporting our enforced "fields inside of sections" block structure.

## Visuals

Peek 2022-07-19 11-18

## Testing Instructions

- Field blocks can be added to a section via drag-and-drop.
- Field blocks cannot be added outside of a section via drag-and-drop.
- Field blocks not supporting `multiple` are disabled when present in a section.

## Pre-review Checklist

-   [ ] Includes unit tests


packages/form-builder/src/blocks/fields/index.js
```diff
@@ -10,13 +10,25 @@ import DonorName from "./donorName";
 const fieldBlocks = [
```
jonwaldstein, line 10: Would like to see this list of blocks broken out into their own files so we can just return a list of field names instead of a really long config.

```
const fieldBlocks = [
  textField,
  donorNameField,
  emailField
]
```

packages/form-builder/src/components/sidebar/secondary.js
```diff
@@ -1,12 +1,15 @@
...
-);
+function Sidebar(props) {
 
-function Sidebar() {
+    const panels = {
+        add: () => <FieldTypesList />,
+        list: () => <BlockListTree />,
+    };
+
+    const PanelContent = panels[props.selected];
```
jonwaldstein, line 12: Can you have this component receive the deconstructed props

`Sidebar({selected})` and specify the prop-type as well.

packages/form-builder/src/components/sidebar/panels/field-types-list.js
```diff
@@ -0,0 +1,77 @@
...
+                        <FieldTypeSectionHeader text={label} />
+                        <BlockTypesList key={name} items={blocks} />
+                    </>
+                );
+            })}
+        </>
+    );
+};
+
+const FieldTypeSectionHeader = ({text}) => {
```
jonwaldstein, line 63: Can you please specify prop-types for this?  Typescript makes this easier but since we're not using that yet let's remember to add prop-types to all components.

jonwaldstein (changes requested):
@kjohnson The drag & drop functionality is great! Only problem is it's not preventing me from dragging a _disabled_ field (even if it has already been added to the form).  Can you please see if we can prevent the drag & drop if a field is disabled/greyed out?

packages/form-builder/src/blocks/fields/index.js
kjohnson: Resolved by https://github.com/impress-org/givewp-next-gen/pull/39/commits/215de048f86d35af32090d7a99cc2f52abdde377

packages/form-builder/src/components/sidebar/secondary.js
kjohnson: @jonwaldstein does this look right? See https://github.com/impress-org/givewp-next-gen/pull/39/commits/e9ad8e34e7ad512fd7c0730be422edf57488da22

packages/form-builder/src/components/sidebar/panels/field-types-list.js
kjohnson: I ended up just in-lining the custom header component, see https://github.com/impress-org/givewp-next-gen/pull/39/commits/def9d9b8fa0a32eba868d83d20c97a5f7e44a318

packages/form-builder/src/components/sidebar/secondary.js
kjohnson:
```js
interface Props {
    selected: 'text'
}

const Sidebar: React.FC<Props> = ({selected}) => { ... }
```

jonwaldstein: @kjohnson I would avoid using React.FC since it implies you also want to accept children.  For this I would just do

```
type PropTypes = {
    selected: string
}

const Sidebar = ({selected}: PropTypes) => {
```

or

```
const Sidebar = ({selected}: { selected: string }) => {
```

Types vs Interfaces are really interested & confusing in typescript.  I generally just use types for components props unless they are extending an existing interface of props or are intended to be extended.

kjohnson: @jonwaldstein cool. Thanks for that. This is actually my first time using types with TypeScript.

Updated. See https://github.com/impress-org/givewp-next-gen/pull/39/commits/86b7b1a72f118bc3dfcf6366db908b1825b41dcf

jonwaldstein (approved):
Nice job!

PR #44, opened 2022-07-26
fix: Disabled blocks are no longer draggable

### LINKED ISSUE #41: Disabled blocks should not be draggable
> [I]t's not preventing me from dragging a disabled field (even if it has already been added to the form) https://github.com/impress-org/givewp-next-gen/pull/39#pullrequestreview-1045639020

Blocks in the `<BlockTypesList />` that are set with `isDisabled` are still draggable.

Source diving Gutenberg shows that the `isDisabled` prop is used to disable the block visually in the list, but does not change the `isDraggable` property - which is inherited from the parent `isDraggable` (at the list level).

kjohnson:
Resolves #41

## Description

This PR forks the `BlockTypesList` component from `@wordpress/block-editor/build/components` so that the `isDraggable` prop accounts for the individual items' `isDisabled` prop.

## Visuals

Peek 2022-07-26 19-06

## Testing Instructions

- Disabled items in the block list should not be draggable.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein (approved):
@kjohnson great work hunting this bug down!  I actually went ahead and updated this fork to use the actual react component src instead of the build file.  That should make this more maintainable and readable without the baggage from the build.

kjohnson:
@jonwaldstein thanks for that change - much better!

PR #49, opened 2022-08-03
Update amount field block

### LINKED ISSUE #48: Update amount field block
## Acceptance Criteria

- [x] Update level button
- [x] Update delete button
- [x] Adjust spacing
- [x] Fix repeater alignment

## Visuals

Desktop - Donation level block(1)

kjohnson:
Resolves #48

## Description

- Updates level button color in canvas (block list)
- Updates level repeater in sidebar
- Updates level repeater delete button
- Updates level repeater add button
- Adds currency prefix to level inputs (hard-coded USD)

## Affects

- Refactors the `donation-amount-levels` field block to shared directory
- Updates application colors to add missing value

## Visuals

| Before |
| --- |
| image |

| After |
| --- |
| image |

## Pre-review Checklist

-   [ ] Includes unit tests


packages/form-builder/src/blocks/fields/amount/inspector/index.js
```diff
@@ -0,0 +1,73 @@
+import {PanelBody} from "@wordpress/components";
+import {__} from "@wordpress/i18n";
+import {InspectorControls} from "@wordpress/block-editor";
+import DeleteButton from "./delete-button";
+import AddButton from "./add-button";
+
+const Inspector = (props) => {
```
jonwaldstein, line 7: Let's always destruct the props whenever possible.  For this we can have `const Inspector = ({attributes, setAttributes}) => {` instead of referencing the props object everytime.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
```diff
@@ -0,0 +1,73 @@
...
+                                                style={{
+                                                    position: 'absolute',
+                                                    height: '100%',
+                                                    display: 'flex',
+                                                    alignItems: 'center',
+                                                    paddingLeft: '13px',
+                                                    fontSize: '.8em',
+                                                }}>
+                                                <spann style={{marginBottom: '1px'}}>$</spann>
+                                            </div>
```
jonwaldstein, line 39: spann! lol

Can we make this currency symbol dynamic?

packages/form-builder/src/blocks/fields/amount/inspector/index.js
```diff
@@ -0,0 +1,73 @@
...
+                                            <input
+                                                style={{
+                                                    width: '140px',
+                                                    padding: '6px 0 6px 30px',
+                                                }}
+                                                value={label}
+                                                onChange={(val) => {
+                                                    const levels = [...props.attributes.levels];
+                                                    levels[index] = val;
+                                                    props.setAttributes({levels: levels});
```
jonwaldstein, line 49: This is an html onChange event so it needs to be 

```
onChange((element) => levels[index] = element.target.value;)
```

packages/form-builder/src/blocks/fields/amount/index.js
```diff
@@ -0,0 +1,37 @@
+import {Icon} from '@wordpress/icons';
+import {__} from "@wordpress/i18n";
+import settings from "../settings";
+
+import Edit from './edit';
+
+const amount = {
+
+    name: 'custom-block-editor/donation-amount-levels', // @todo Rename this block.
+
```
jonwaldstein, line 10: Has this todo been completed?

packages/form-builder/src/blocks/fields/amount/inspector/index.js
```diff
@@ -0,0 +1,73 @@
...
+                                            props.setAttributes({levels: props.attributes.levels.slice()});
+                                        }} />
+                                    </li>
+                                );
+                            })
+                        }
+                    </ul>
+                )}
+                <AddButton onClick={() => {
+                    const levels = [...props.attributes.levels];
```
jonwaldstein, line 64: FYI 
`const levels = [...props.attributes.levels];` 
is the same thing as 
`const levels = props.attributes.levels;`

jonwaldstein (changes requested):
@kjohnson lookin good!  A few suggestions and the level value one is causing the block to break when changing a value currently.

Also just wanted to point out a few React dev consistencies i'd like to see us improve upon as this app continues to grow:
1. Destruct props at the component function level whenever possible.
2. I'd like to see component file naming conventions as `/MyComponent.(jsx|tsx)` and `/myNonComponent.(js|ts)`. Linting machines and bundlers like to see that jsx file extension.  It's also easier to differentiate component files from utility functions along with the capitalizing naming convention outlined here.
3. Start using more typescript.  At this point I don't see any reason not to since it makes our front-end code more stable with types.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: It looks like we can add (or enable) this as an `eslint` rule, see https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md

jonwaldstein: Yeah, unfortunately we're not really using eslint anymore (maybe react scripts does?). It's something i've realized recently that we're missing in our react dev and should start using again.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: I'm cloning the object here, which seems to be required when passing it to `setAttributes()`.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: Ah, I missed that when refactoring away from `<TextControl/>`. Resolved in https://github.com/impress-org/givewp-next-gen/pull/49/commits/6070228495fbe8f755fce7bdead42aa88be9adf3.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: I'm using eslint with create-react-app (but the linked plugin doesn't work with that configuration). For now I'm happy to fix it manually. Resolved in https://github.com/impress-org/givewp-next-gen/pull/49/commits/6070228495fbe8f755fce7bdead42aa88be9adf3.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: Tag name corrected, see https://github.com/impress-org/givewp-next-gen/pull/49/commits/6e32f27eb950235bec2c2d403c2ccf54334c2c49.

kjohnson: I haven't addressed currency in the builder yet (still need to think about currency switcher). Is that something that can wait, for now?

packages/form-builder/src/blocks/fields/amount/index.js
kjohnson: I left the todo because I hadn't decided on a new name yet, but I'll being doing a pass soon to rename all the block namespaces and will do it then.

packages/form-builder/src/blocks/fields/amount/inspector/index.js
jonwaldstein: Since the currency symbol is dynamic on the front-end, let's make it appear so here too.  We can just use the global currency option for now until we introduce per-form control.  Here's what's happening on the front-end that should be virtually the same here:

Screen Shot 2022-08-08 at 9 13 19 AM

Screen Shot 2022-08-08 at 9 13 35 AM

packages/form-builder/src/blocks/fields/amount/inspector/index.js
jonwaldstein: Sorry, I was just pointing out that spreading a single array inside an empty array is the same as referencing the array in the first place, so you don't need to do the spreading here.

kjohnson: It was breaking if I don't spread the array into a new array. I think there is an issue with mutability at play. If I update the attribute value directly, passing it to `setAttributes()` does not update the attribute. However, if I spread the value into a new object then it works.

```js
const a = []
const b = a // assign by reference
const c = [ ...a ] // assign by value

console.log( a === b ) // true
console.log( a === c ) // false
```

JasonTheAdams: Just because I know y'all would care, technically _everything_ in JavaScript is passed by value, including objects. Objects, however, are a set of references — a reference of references. So when passing an object, it creates a copy of reference. This means that the underlying references are the same, which gives it the feeling of being by reference. This evaluation strategy is called [Call by sharing](https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_sharing).

So when you use `Object.assign()` or spreading, the underlying object references (keep in mind that arrays are objects in JS) are resolved to pass by value. BUT! This is only true for top-level values. Call by sharing takes effect again for nested objects:

```js
let foo = { a: 1, b: 2, c: { d: 3 } };
let bar = { ...foo };

foo.a = 10;
foo.c.e = 4;

bar.a; // 1
bar.c.e; // 4
```

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: @jonwaldstein I localized the currency from the Give settings and implemented `react-currency-input-field`. See https://github.com/impress-org/givewp-next-gen/pull/49/commits/500f5eb22c5bcc75ac7b3ed119604bf74b342ffd.

jonwaldstein: @kjohnson I refactored this a little bit here 72c99d51 just to lean on the `react-currency-input-field` library more and let them handle the currency symbol.  We're really trying to move away from hard-coding the symbols (fyi our currency settings are storing hard-coded symbols) and let js/php generate the locale & symbols based on the currency.

jonwaldstein (approved):
@kjohnson looking good!

packages/form-builder/src/blocks/fields/amount/inspector/index.js
kjohnson: Sure that works for me.

PR #53, opened 2022-08-17
Feature: Add Payment Gateways block and make section act normal

JasonTheAdams:
## Description

Up to this point the "payment details" section was special. It didn't actually conform to what the Form Builder did, as it was uniquely identified and rendered in the form code. In part, this was because there was no Payment Gateway block.

This PR introduces the Payment Gateway block and template, and normalizes how the section functions. Very likely we'll want to adjust how the template works in the future, as it currently relies on hooks that make it hard to override, but this is good enough for now. It's more important to get the section working normally so there isn't a disconnect between itself and the Form Builder.

We are still adding a couple hidden fields to that section. We'll want to move these at some point and determine where/how we'd like to add those critical fields.

Lastly, I took a moment to clean up some Classic template styles and make the Donation Summary look more like the original design.

## Affects

The payment details section as a whole, really. Also, stylistically, I tweaked all sections for the Classic template to improve spacing between fields.

## Visuals

image

## Testing Instructions

Check that adding fields to the payments section now works, as well as moving things around.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/Framework/FieldsAPI/PaymentGateways.php
+++ b/src/Framework/FieldsAPI/PaymentGateways.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Give\Framework\FieldsAPI;
+
+class PaymentGateways extends Field
+{
+    public function getNodeType(): string
+    {
+        return 'layout';
+    }
+
+    public function getType(): string
+    {
+        return 'gateways';
+    }
+}
--- a/src/NextGen/DonationForm/Blocks/DonationFormBlock/Block.php
+++ b/src/NextGen/DonationForm/Blocks/DonationFormBlock/Block.php
@@ -12,6 +12,7 @@ use Give\Framework\FieldsAPI\Exceptions\TypeNotSupported;
 use Give\Framework\FieldsAPI\Form;
 use Give\Framework\FieldsAPI\Hidden;
 use Give\Framework\FieldsAPI\Name;
+use Give\Framework\FieldsAPI\PaymentGateways;
 use Give\Framework\FieldsAPI\Radio;
 use Give\Framework\FieldsAPI\Section;
 use Give\Framework\FieldsAPI\Select;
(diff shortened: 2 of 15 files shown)


src/Framework/FieldsAPI/PaymentGateways.php
```diff
@@ -0,0 +1,18 @@
...
+
+declare(strict_types=1);
+
+namespace Give\Framework\FieldsAPI;
+
+class PaymentGateways extends Field
+{
+    public function getNodeType(): string
+    {
+        return 'layout';
```
kjohnson, line 11: This is usually only set on the base, or abstract, node class and isn't needed here unless there is some specific use case.

src/Framework/FieldsAPI/PaymentGateways.php
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    public function getType(): string
+    {
+        return 'gateways';
```
kjohnson, line 16: This method is inherited from the base node type class, which expects a class constant of `TYPE` where the type should be defined in each class.

kjohnson (changes requested):
Functionality looks good. Just a couple of changes to the Fields API node definition.

src/Framework/FieldsAPI/PaymentGateways.php
JasonTheAdams: I did this simply because I wasn't sure how to categorize the `Gateways` node. In the form rendering engine I made it a layout, but perhaps that's not accurate. It technically is a radio field. So perhaps it's a Field?

src/Framework/FieldsAPI/PaymentGateways.php
JasonTheAdams: Ah! Ok. Resolved in eb3fb7b

src/Framework/FieldsAPI/PaymentGateways.php
kjohnson: I would say Field, which is already set by the base Field class.

JasonTheAdams: Thanks for bringing this up, @kjohnson! I took another stab at this and I'm much happier with it. I've cleaned up quite a bit and now this actually just acts like a fancy radio field. Tentatively resolved in 2566434

Issue #24, opened 2022-07-06
Field names should be unique and predictable


kjohnson:
Each block has a generated UUID, but this is not developer friendly when we get to the Fields API. Core fields will have names by convention, such as `donor-name`, etc, but custom fields will need a unique `name` attribute for reference in the Fields API.

## Acceptance Criteria

- [x] Core fields are referenced by convention based names
- [ ] Custom field name are unique

PR #55, opened 2022-08-18
Feature: add paragraph element

JasonTheAdams:
## Description

This adds a paragraph element to the Form Builder! It allows folks to add simple text anywhere in any form. I looked into text formatting, and even added support for the core bold, italics, and link formats, but we apparently will need to [register format types](https://webomnizz.com/gutenberg-editor-add-button-to-the-toolbar/) in our app to get those to work properly. The [core types](https://github.com/WordPress/gutenberg/blob/0638a8c52199381080abb809fd5ccd137db5bb45/packages/format-library/src/bold/index.js) look easily available. I think it's fine if we add those in a subsequent PR.

## Affects

Didn't really change too much, just added the element.

## Testing Instructions

Use the block (or a few) in the Form Builder and check it out!

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/packages/form-builder/src/blocks/elements/paragraph.tsx
+++ b/packages/form-builder/src/blocks/elements/paragraph.tsx
@@ -1,0 +1,60 @@
+import {__} from '@wordpress/i18n';
+import {RichText, RichTextToolbarButton} from '@wordpress/block-editor';
+import {Icon} from '@wordpress/icons';
+
+const paragraph = {
+    name: 'custom-block-editor/paragraph',
+    category: 'content',
+    settings: {
+        title: __('Paragraph', 'custom-block-editor'),
+        supports: {
+            html: false,
+            multiple: true,
+        },
+        attributes: {
+            content: {
+                type: 'string',
+                source: 'html',
+                default: '',
+            },
+        },
+        icon: () => (
+            <Icon
+                icon={
+                    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+                        <line x1="14.917" y1="20" x2="14.917" y2="4.88889" stroke="#1E1E1E" strokeWidth="1.5" />
+                        <line x1="10.4727" y1="20" x2="10.4727" y2="4.88889" stroke="#1E1E1E" strokeWidth="1.5" />
+                        <line x1="18.333" y1="4.75" x2="9.44412" y2="4.75" stroke="#1E1E1E" strokeWidth="1.5" />
+                        <path
+                            d="M9.13889 8.88889V12.96C7.21109 12.6071 5.75 10.9186 5.75 8.88889C5.75 6.85914 7.21109 5.17065 9.13889 4.81778V8.88889Z"
+                            fill="#1E1E1E"
+                            stroke="#1E1E1E"
+                            strokeWidth="1.5"
+                        />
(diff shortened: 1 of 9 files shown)


JasonTheAdams:
@kjohnson Something I'm curious what you think about. In making this I realized that paragraphs don't really have a "name". When registering it with the Field API it requires a name... but there really isn't one. To get past it I just generated a random hash, but I don't think that's great. We just don't really need a name as nothing is saved, but the is also useful for retrieving the node within the tree — so random isn't helpful. Thoughts?

kjohnson:
@JasonTheAdams on the roadmap I plan to auto-generate names for blocks, see https://github.com/impress-org/givewp-next-gen/issues/24. This will provide a reference to the generated name in the block settings so that they can use the Fields API to update, move, or remove the node.

That said, a random hash is good enough for now 👍🏻 .

kjohnson:
Of course, I was holding off thinking your weren't yet ready - but I'll jump on this now.

src/NextGen/DonationForm/Blocks/DonationFormBlock/app/templates/elements/Paragraph.tsx
```diff
@@ -0,0 +1,9 @@
+import {ElementProps} from '@givewp/forms/propTypes';
+
+interface ParagraphProps extends ElementProps {
+    content: string;
+}
+
+export default function Paragraph({content}: ParagraphProps) {
+    return <p>{content}</p>;
```
kjohnson, line 8: I wonder if it would simplify things to re-use the `html` element, since the only difference here is a wrapping tag. Not wanting to hold up this PR but something to consider/discuss.

packages/form-builder/src/blocks/elements/paragraph.tsx
```diff
@@ -0,0 +1,60 @@
...
+                    </svg>
+                }
+            />
+        ),
+        edit: ({attributes, setAttributes}) => {
+            const {content} = attributes;
+            return (
+                <>
+                    <div>
+                        <RichText
```
kjohnson, line 43: I'm seeing a `contenteditable` bug in Firefox, which is not present in the Gutenberg Paragraph block. At some point we should take a closer look at the official implementation for these sort of considerations.

kjohnson (approved):
Tested functionality in the form builder and the form template - ✅.

[I have some questions, a couple of suggestions](https://user-images.githubusercontent.com/10858303/185440537-6d97c438-04cb-4dd3-ba77-38e16f54153b.gif) - but those aren't blocking for the PR.

src/NextGen/DonationForm/Blocks/DonationFormBlock/app/templates/elements/Paragraph.tsx
JasonTheAdams: Yeah, I thought about that. Ultimately, I want different rules for what tags are supported. I'll probably use https://interweave.dev/

packages/form-builder/src/blocks/elements/paragraph.tsx
JasonTheAdams: Normally, I'd pause to address this, but I'll make a note and circle back to it in a subsequent PR after the Town Hall. We just won't use Firefox.

PR #64, opened 2022-09-21
Refactor: make inputs readOnly

### LINKED ISSUE #62: Form builder label settings UX should be improved
## Details

The functionality of form label settings in the builder should be improved for a better user experience.  Currently there is a disconnect between settings on the field in the canvas and in the sidebar.  It's not clear where the setting should be set because both are interactive. People should be able to to click on a label in the canvas and change it there.  The field should also be read-only inside the builder to make it less confusing.

## Acceptance Criteria

- [ ] Labels can be changed from the canvas
- [x] Form inputs are read-only

jonwaldstein:
Resolves #62

## Description

This PR adds the `readOnly` property to field inputs to avoid confusion about field label and placeholder settings.  With this change, the intuitive way to edit those settings are in the block settings themselves.

Along with this update, the main entry point to the form builder was converted to typescript so that block configurations are typed.

## Affects

Form builder field inputs are now restricted to readOnly.

## Testing Instructions

Compile assets and interact with fields in the form builder.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@kjohnson this is ready for review.  cc: @JasonTheAdams 

FYI - we previously talked about labels being _editable_ within the canvas.  I realized that would be _a lot_ more work to accomplish because we would have to fork the Wordpress `<TextControl>` component which we are using for all the fields right now. I did attempt to add a separate editable label alongside the `TextControl` but there was some strange side affects around field focusing and settings. 

So for this PR I decided to limit the scope to just making fields `readOnly` which still improves the UX because the field settings are now more intentional and should avoid confusion.  

If we really feel having labels being _editable_ within the canvas is the best experience - then we can explore making our own wrapper around the `TextControl` component.

JasonTheAdams:
I'm cool with this scope adjustment. We can tackle making the labels editable another time. It's not critical, and adjusting the label in the settings has immediate feedback in the canvas, which is good enough.

kjohnson (approved):
Looks good to me. And thanks for moving the file to TypeScript during the refactor.

PR #65, opened 2022-09-27
Feature: Add auto-generated, unique field names

### LINKED ISSUE #24: Field names should be unique and predictable

kjohnson:
Resolves #24

## Description

This PR adds support for field names that are unique and predictable, auto-generating incremented suggestions when a duplicate is found, and originally derived from the field's label.

## Todo

- [x] Validate for unique
- [x] Validate for required
- [x] Validate for format (slug)
- [x] Update Block to Fields API conversion to support field names

## Visuals

image

## Testing Instructions

- Create a new, custom Text Field and set the Field Name attribute to `tester`.
- Create a another new, custom Text Field and set the Field Name attribute to `tester`.
  - The Field Name attribute should be updated to `tester-1` on blur.
- Create a another new, custom Text Field and set the Field Name attribute to `tester`.
  - The Field Name attribute should be updated to `tester-2` on blur.

## Pre-review Checklist

-   [x] Includes unit tests


kjohnson:
**Thinking Out Loud**

I've been thinking through ways to extract this functionality for re-use, but don't yet have a solid direction.

And this discussion doesn't need to hold-up this current implementation, which I think is sufficient for now.

---

One idea is to use `@wordpress/compose` to, essentially, create traits or concerns that can be decorated onto a component. This would make it easy for a developer to apply these "traits" to a field/block.

```jsx
export default compose(
    withFieldName,
    withRequiredInput,
)( CustomFieldComponent );
```

This would also allow each "trait" to co-locate properties, functions, and JSX that are otherwise spread out across the component.

For example, the underlying field name logic is encapsulated in a hook (`useFieldNames`), but this still requires that each component using this hook connect the returned validator to the `<InspectorControls>`'s corresponding `<TextControl>`. The logic and the control are in two different parts of the field's component.

By contrast, a `withFieldName` Higher-Order Component would encapsulate the logic and the control  in a wrapper around the underlying field component.

This is how Gutenberg shared functionality for Instance IDs, keyboard shortcuts, and reduced motion, which are used across multiple components.

packages/form-builder/src/blocks/fields/settings.js
```diff
@@ -1,6 +1,58 @@
...
 import {__} from "@wordpress/i18n";
+import {noop} from "lodash";
+
+const Edit = function (props) {
+
+    const {
+        attributes: {label, placeholder, isRequired, options},
+        setAttributes,
+        onLabelTextControlBlur = noop
+    } = props;
```
jonwaldstein, line 12: Let's make sure to destruct the props inline within the component function so they're more explicit.

packages/form-builder/src/hooks/useFieldNames.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const getFieldNameSuggestion = (name, names) => {
+    const [ prefix ] = name.split(/^(.*)-([0-9]*)$/g).filter(Boolean)
+    const similarFieldNames = names.filter(fieldName => fieldName.startsWith(prefix));
+    const increments = similarFieldNames.flatMap(fieldName => fieldName.split(/^.*-([0-9]*)$/g).filter(Number) ) || [ 0 ]
+    const nextIncrement = increments.length ? Math.max(...increments) + 1 : 1
+    return `${prefix}-${nextIncrement}`;
+}
+
+export const flattenBlocks = (block) => [block, ...block.innerBlocks.flatMap(flattenBlocks)]
```
jonwaldstein, line 15: This looks nifty, can you please write some detailed descriptions of everything going on in this file for our future selves?

packages/form-builder/src/blocks/fields/field.js
```diff
@@ -1,6 +1,73 @@
 import {Icon} from '@wordpress/icons';
 import {__} from "@wordpress/i18n";
 import settings from "./settings";
+import {useFieldNames} from "../../hooks";
+import {InspectorAdvancedControls} from "@wordpress/block-editor";
+import {ExternalLink, PanelRow, TextControl} from "@wordpress/components";
+
+const Edit = (props) => {
```
jonwaldstein, line 8: Again, can we please destruct the props inline here.

packages/form-builder/src/blocks/fields/field.js
```diff
@@ -1,6 +1,73 @@
...
+    const ParentEdit = settings.edit
```
jonwaldstein, line 9: Can you explain what's up with the `ParentEdit` thing?  

As a side note, we always want to avoid creating/instantiating variables, components _inside_ a component to avoid side effects and re-renders.  In this case you would want to define this before `const Edit`

packages/form-builder/src/blocks/fields/field.js
```diff
@@ -1,6 +1,73 @@
...
+    const {attributes: {fieldName, label}, setAttributes} = props;
+
+    const validateFieldName = useFieldNames()
+
+    const slugify = (value) => {
+        return value.toLowerCase()
+                    .replace(/\s|_/g, '-') // Replace spaces and underscores with dashes
+                    .replace(/[^a-zA-Z\d\s-]/g, '') // Replace non-alphanumeric characters (other than dashes)
+                    .replace(/-$/g, '') // Remove trailing dash
+    }
```
jonwaldstein, line 20: This can be extracted outside the component, let's throw it in the common folder and pull it in here.

jonwaldstein (changes requested):
@kjohnson thanks for putting this together, it seems to be working properly.  Just had a few suggestions

packages/form-builder/src/blocks/fields/field.js
jonwaldstein: Resolved c072dee6

packages/form-builder/src/blocks/fields/settings.js
jonwaldstein: Resolved 8778f1ed

packages/form-builder/src/blocks/fields/field.js
jonwaldstein: resolved c072dee6

packages/form-builder/src/blocks/fields/field.js
jonwaldstein: resolved f5a47a87

jonwaldstein (approved):

PR #71, opened 2022-10-11
Feature: Dynamic, more powerful front-end validation

JasonTheAdams:
## Description

The primary goal of this PR is to receive the validation rules from the Field API and apply them to the front-end validation by converting them to [Joi](https://joi.dev) rules and using the React Hook Form resolver.

Next, it streamlines labels and errors for fields by adding two components to the FieldProps
- `Label` will automatically include an (accessible) asterisk if the field is required
- `ErrorMessage` will automatically display if a field has an error message (and show nothing if not)

While I was doing this, I also noticed a bunch of code does similar things — recursively walks/maps/reduces nodes in order to serve some purpose. Each place handled this on its own and in its own way. So I thought it would be useful to beef out the `groups.ts` file with a bunch of useful functions for traversing nodes. The group nodes themselves also have new methods for walking, looping, and reducing their nodes.

## Affects

How front-end validation works

## Testing Instructions

Mostly, make sure that the validations still work and that nothing is broken. Otherwise, this is largely a code review.

## Pre-review Checklist

-   [ ] Includes unit tests


JasonTheAdams:
This bad boy is ready for review, @jonwaldstein!

src/Framework/FieldsAPI/Amount.php
```diff
@@ -18,6 +18,15 @@ class Amount extends Field
      */
     protected $allowCustomAmount = false;
 
+    public function __construct($name)
+    {
+        parent::__construct($name);
+
+        $this->validationRules->rule('numeric', true);
+        $this->validationRules->rule('integer', true);
+        $this->required();
```
JasonTheAdams, line 27: I was really on the fence about this, @jonwaldstein, and would like your thoughts. The alternative is to apply all validation rules during the conversion process.

jonwaldstein: Ah yeah, I was wondering where the validation rules for the amount field went.  I think for consistency, I would prefer these to be added during the conversion process - especially required.  If we start using this field in other places and it's always going to be the same validation rules then maybe i can see this approach valuable as an abstraction - but so far other fields don't have them baked in.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/fields/Email.tsx
```diff
@@ -1,11 +1,15 @@
 import type {FieldProps} from '@givewp/forms/propTypes';
 
-export default function Email({label, placeholder, fieldError, inputProps}: FieldProps) {
+export default function Email({label, requiredLabel, placeholder, fieldError, inputProps}: FieldProps) {
     return (
         <label>
-            <span>{label}</span>
-            <input type="email" aria-invalid={fieldError ? 'true' : 'false'}
-                   placeholder={placeholder} {...inputProps} />
+            <span>{requiredLabel}</span>
```
jonwaldstein, line 6: Where does the original `label` prop go now?  I see this change to the other field components as well.  Here it's leftover but not applied.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/ConvertFieldAPIRulesToJoi.ts
```diff
@@ -0,0 +1,83 @@
...
+import {Field, Form} from '@givewp/forms/types';
+import {__, sprintf} from '@wordpress/i18n';
+import {reduceFields} from './groups';
+
+const requiredMessage = sprintf(
+    /* translators: base error message */
+    __('%s is required.', 'give`'),
+    `{#label}`
+);
+
```
jonwaldstein, line 11: I noticed the amount field is not working properly, the label should be `"Donation Amount"` which is hidden by the template.  I will also point out that it's a little strange to refer to a field that doesn't a visible label here but I guess it's within context so not a big deal.

Screen Shot 2022-10-13 at 2 25 55 PM

jonwaldstein (changes requested):
This is a great update! Just had a few comment/questions along with the thread about Amount field validation.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/DonationFormBlockApp.tsx
```diff
@@ -3,27 +3,33 @@ import getDefaultValuesFromSections from './utilities/getDefaultValuesFromSectio
...
 import getWindowData from './utilities/getWindowData';
+import prepareFormData from './utilities/PrepareFormData';
+import getJoiRulesForForm from './utilities/ConvertFieldAPIRulesToJoi';
 
 /**
  * Get data from the server
  */
 const {form} = getWindowData();
 
+prepareFormData(form);
```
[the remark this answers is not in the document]
JasonTheAdams: Memory preservation for such a large object. I didn't see any reason to do this immutably.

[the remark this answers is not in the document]
jonwaldstein: Oh I see, it's just assigning the `requiredLabel` property to fields.  What else do you anticipate doing in here? Wondering if this function should stay kind of generic or be named more specific since at first glance it's not clear what exactly it's  _preparing_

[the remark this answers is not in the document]
JasonTheAdams: This now also adds node handling functions to the form and its groups in 9ad0deb.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/form/Form.tsx
```diff
@@ -132,6 +118,7 @@ export default function Form({defaultValues, sections}: PropTypes) {
 type PropTypes = {
     sections: Section[];
     defaultValues: object;
+    schema: ObjectSchema;
```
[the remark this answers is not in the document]
JasonTheAdams: Good idea. Resolved in 392bb9b.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/fields/Email.tsx
JasonTheAdams: I cleaned this, and some other stuff, up in c58f88f.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/ConvertFieldAPIRulesToJoi.ts
JasonTheAdams: I think that the context is fine. I also updated this to say that it's required like the other fields in 47f0b1e.

src/Framework/FieldsAPI/Amount.php
JasonTheAdams: I can't move them all over to the conversion, yet, as the validationRules property is protected in core. I want to think through that a bit more, too. So I've at least put required back in f45d2f7.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/fields/Text.tsx
```diff
@@ -1,9 +1,9 @@
 import type {FieldProps} from '@givewp/forms/propTypes';
 
-export default function Text({label, placeholder, fieldError, inputProps, validationRules}: FieldProps) {
+export default function Text({requiredLabel, placeholder, fieldError, inputProps}: FieldProps) {
     return (
         <label>
-            <span>{label}</span>
+            <span>{requiredLabel}</span>
```
jonwaldstein, line 6: This `requiredLabel` prop is confusing me.  What if a field is not required? wouldn't we need a conditional here to display the `label` string instead of the `requiredLabel` jsx?

JasonTheAdams: Yeah, I agree that it's a little weird. The problem I was trying to solve was introducing the accessible asterisk _without_ having to make every single template file handle that itself. I thought about simply making a `<RequiredLabel />` component that template components could use, but realized we'd need to make it easily accessible to 3rd parties, and it's adding even more mental complexity for something simple. I then thought to just update the label itself, so folks don't worry about it at all! But then I realized it's important to preserve the label as a simple string for things like error messages. So then I arrived at making a new label prop that template components would simply use in place of the label and just not worry about it.

Perhaps this could be flipped on its head. I could take the original label, assign it to a `Field.rawLabel` (or something like) property, and then update `Field.label`. Then the template components would just use `label` as they have. We'd then have `rawLabel` to use internally for things like messages.

jonwaldstein: I hear you. I think it's useful to generate common html like this for templates I just want to keep the compositional templating in tact. Here's an example of what I would be happier to see inside a field.

`<span>{isRequired ? labelWithRequiredIndicator : label}</span>`

JasonTheAdams: I'm using this to think through this one last time, @jonwaldstein :

### Method 1: Replace the Label
- Update the `Field.label` so it includes the indicator when necessary.
- Add a `Field.plainLabel` which is always the simple, text version of the label (no indicator)

**Pros**
- Templates do nothing special to display the requirement indicator
- Requirement remains a validation concept and isn't used in the template

**Cons**
- Template has to know about the `Field.plainLabel` if they for some reason want to do something with the label and not show the indicator

### Method 2: Include a new labelWithRequiredIndicator property
- Add a `Field.isRequired` property as a shorthand for `Field.validationRules.required` _(optional)_
- Add a `Field.labelWithRequiredIndicator` property which templates can use

**Pros**
- Whether or not the plain label or label with indicator is used is clear in the template and up to the template
- The `Field.label` is predictably always just the plain label string

**Cons**
- Every template has to include the condition as to which form of the label to display based on requirement
- We risk template overrides not realizing they need to do this ❗
- Field templates must think about validation in order to render properly

---

Ok. That really helped! I'm really leaning towards Method 1 at this point. Running various scenarios through my head, I can't actually think of a reason (or at least common one) wherein someone would want to hide the indicator when rendering the field. If they do, then they're probably overriding, and frankly it would be simpler to just hide the indicators via CSS (which would keep them for accessibility). Method 2 ends up putting the onus on each template (and override) to include the same condition. As noted, I think this is more likely to create scenarios wherein folks get confused why the indicator isn't showing up — not realizing its up to them. It also begs the question as to whether `Field.labelWithRequiredIndicator` is null or without the indicator if the field isn't required, which opens up more room for issues/confusion.

Unless you can think of a serious counterpoint (which I'm open to) I'm going to move forward with Method 1. In any case, it's not what I started with, so thank you for thinking through this with me!

jonwaldstein: Thanks for writing this down. 

So there's 1 main scenario that I think could warrant another look at method 2 or even another solution that would satisfy them.

**Scenario**
Template wants to change the asterisk to the word "required".  
   - example: `First name (required)` vs `First name *`

The asterisk and word 'required' are both recommended ways of providing accessible required indicators for a label.  I think it would be valuable for templates to have control over which they prefer to display instead of being locked into one or the other.

Also, I think this discussion should include field errors as well.  Notice how we are handling errors in fields by using a conditional for `fieldError` and then displaying some html.  Whatever direction we decide for labels should reflect other conditional template logic like field errors.

_As I was writing this I just came up with this idea..._

Here's one more method to throw in the mix:

**Method 3: extract common field props into their own template (derived from FieldNode)** 

With this method we could still have the simplicity of **method 1** where the template would just display {label} and {fieldError} but with the flexibility of overriding the html for all fields. 🤯 

```
function Label({label, required}) {
    return (
        <>
            {label}{' '}
            {required && (
                <span className="givewp-field-required" title="Field Required">
                    *
                </span>
            )}
        </>
    );
}

function FieldError({error}) {
    return (
        error && (
            <div className="error-message">
                <p role="alert">{error}</p>
            </div>
        )
    );
}

template = {
    fieldProps: {
        label: Label,
        fieldError: FieldError
    }
}
```

JasonTheAdams: We've done it. 😎

JasonTheAdams:
Alright, @jonwaldstein, I've done some things. I'm actually pretty pleased with a couple unexpected improvements. Here's what we've got:
- Fields have a `Label` component which handles requirements and can be overridden as a template
- Fields have a `ErrorMessage` component which conditionally displays the error and can be overridden as a template
- I reworked how Groups handle internal fields (more below)
- Some TypeScript magic to make groups and everything typed

**Group changes:**
In order to pass the `Label` and `ErrorMessage` components to groups, I realized I needed to build more things in the `GroupNode` component. I also need to add more `GroupProps`. After going down a road of just adding more properties, I had a "wait, what?" moment. I realized we were passing down only the input props to each field, then thought... why?

So instead of passing down only the field's input props, I decided to pass down the `FieldProps` for every single field! I then consolidated how fields are registered and their props built for both independent Fields and Group Fields. If you look at the `Name` component you'll see how it made that group waaaaay simpler. I'm pretty darned pleased with that. Fun bonus.

Please give this bad boy another review. I'm feeling really good that we've landed on some solid improvements.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/registerFieldAndBuildProps.tsx
```diff
@@ -0,0 +1,24 @@
+import {Field} from '@givewp/forms/types';
+import {FieldProps} from '@givewp/forms/propTypes';
+import {UseFormReturn} from 'react-hook-form';
+import buildRegisterValidationOptions from './buildRegisterValidationOptions';
+import {getFieldErrorTemplate, getFieldLabelTemplate} from '../templates';
+import getErrorByFieldName from './getErrorByFieldName';
+
+export default function registerFieldAndBuildProps(
+    field: Field,
+    register: UseFormReturn['register'],
```
JasonTheAdams, line 10: @jonwaldstein @kjohnson This was an interesting find. I wanted the type to be the same type as another type's property. Apparently this is how you do that.

JasonTheAdams:
@jonwaldstein This is ready for you to review, again. I could also use your hand on the merge conflict as I'm not sure what changed there and which direction it should go.

jonwaldstein (approved):
Dynamic front-end validation... 🤯 Wow, this is such a game changer. Excellent work!

PR #75, opened 2022-10-28
Refactor: donation form now only loads scripts and styles that it needs

jonwaldstein:
## Description

While working on the design preview in the form builder, I discovered a script conflict being loaded from give core.  The real problem is that core scripts _were_ being loaded in the iframe (along with all kinds of other scripts from add-ons). The NextGen donation form has no business loading any other scripts than the ones it enqueues through our framework.  This PR rearranges our script loading process within the donation form to make the order and loading of scripts predictable and specific to our form.

The process was actually a lot simpler than I originally thought.  Instead of calling `wp_head()` and having to dequeue a bunch of stuff we can break out the printing into specific parts...

**This is the order of loading:**
 1. Enqueue global styles from WP using `wp_enqueue_global_styles()`.
    - This ensures template compatibility with global WP css variables as needed. Loads before our templates, so they can use things like global font-family, etc.  Also supports theme.js for FSE.
  2. Enqueue our donation form specific scripts & styles.
    - We will let WP handle the actual printing depending on how they were enqueued.
  3. Call the specific WP functions `wp_print_styles()` and `wp_print_head_scripts()`
    - This will only print the styles and scripts that are enqueued within our route - so we don't have to dequeue a bunch of stuff.
  4. Manually echo our window data and root div for our React app to consume
  5. Finally, call the specific WP function `wp_print_footer_scripts()`
    - This will only print the footer scripts that are enqueued within our route.

## Affects

This affects the donation form but in a good way.  The form only loads the scripts it needs now. No more unused external scripts.

## Visuals

Screen Shot 2022-10-28 at 11 55 32 AM

## Testing Instructions

Just make sure a next gen form look and works the same.  There should be nothing unusual about it.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/NextGen/DonationForm/Controllers/DonationFormViewController.php
+++ b/src/NextGen/DonationForm/Controllers/DonationFormViewController.php
@@ -40,10 +40,15 @@ class DonationFormViewController {
     private function head()
     {
         global $wp_scripts, $wp_styles;
-        wp_head();
+        add_action('wp_print_scripts', function () {
+            wp_dequeue_script('give');
+            wp_dequeue_script('give_recurring_script');
+        });
 
-        $wp_styles->dequeue(array_column($wp_styles->registered,'handle'));
-        $wp_scripts->dequeue(array_column($wp_scripts->registered,'handle'));
+        $wp_styles->dequeue(array_column($wp_styles->registered, 'handle'));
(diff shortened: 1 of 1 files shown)


jonwaldstein:
@kjohnson this is ready for review.  This is essentially blocking design previews, so we'll need to merge this before opening a PR for that.

PR #78, opened 2022-11-01
Refactor: form builder settings are now setup as a store with reducer

### LINKED ISSUE #17: Form settings should use a reducer
Looking at these settings more, I think it would actually make sense to break them out into their own nested objects based on the group of settings to simplify data shape & naming conventions.

```
formSettings {
  donationGoal {
    enable,
    autoClose,
    amount,
    format
  }
}
```

Then with the reducer you could do something like:

```
dispatch(updateDonationGoalSettings({enable: true}))
```

_Originally posted by @jonwaldstein in https://github.com/impress-org/givewp-next-gen/pull/10#discussion_r894834743_

jonwaldstein:
Resolves #17

## Description

As the form builder application continues to grow, it's necessary to think through our data flow as it relates to all the new features we plan on implementing.  This PR replaces the existing state management approach with a reducer powered approach. This will give us a lot more flexibility in the future along with performance benefits.

**Provider**
- FormSettingsProvider

This is our Form's new store/provider.  It uses the React hook useReducer to store and persist state under the hood.  In order to update the state a "dispatch" method will be provided from the reducer.   This provides a lot of flexibility and performance benefits.

**Hooks**
- useFormSettings: convenient way of retrieving all the settings from the store as readOnly
- useFormSettingsDispatch: convenient way of retrieving the "dispatch" function from our provider

```
const FormTitleSettings = () => {
    const {formTitle} = useFormSettings();
    const dispatch = useFormSettingsDispatch();

    return (
        <PanelBody>
            <PanelRow>
                <TextControl
                    label={__('Form Title')}
                    value={formTitle}
                    onChange={(formTitle) => dispatch(setFormSettings({formTitle}))}
                />
            </PanelRow>
        </PanelBody>
    );
};
```

Notice we are dispatching `setFormSettings()` - this is just a declarative abstraction for something like this:
```
dispatch({type: 'update_settings', settings: {formTitle})
```

Creating these reducer dispatch functions make updating nested object much easier.  In the future I anticipate having breaking out the form settings into nested objects that describe themselves with matching functions.  Think design/template settings, goal settings, addon settings, etc.

## Affects

The form builder application's front-end state management.

## Visuals

There is no visual change.

## Testing Instructions

Just make sure everything still works as expected.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1203272136087566

### DIFF
--- a/packages/form-builder/src/App.js
+++ b/packages/form-builder/src/App.js
@@ -1,83 +1,39 @@
-import React, {useState} from 'react';
+import React from 'react';
 
 import {ShortcutProvider} from '@wordpress/keyboard-shortcuts';
-import {BlockEditorProvider, BlockInspector} from '@wordpress/block-editor';
-import {Popover, SlotFillProvider} from '@wordpress/components';
-import {InterfaceSkeleton} from "@wordpress/interface";
-
-import HeaderContainer from './containers/HeaderContainer';
-import {SecondarySidebar, Sidebar} from './components';
-import {defaultFormSettings, FormSettingsProvider} from './settings/context';
+import BlockEditorContainer from './containers/BlockEditorContainer.tsx';
+import {FormSettingsProvider} from './stores/form-settings/index.tsx';
 import {Storage} from './common';
-
-import {useToggleState} from "./hooks";
 
 import '@wordpress/components/build-style/style.css';
 import '@wordpress/block-editor/build-style/style.css';
 
 import './App.scss';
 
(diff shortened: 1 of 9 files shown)


jonwaldstein:
@kjohnson this is ready for review.  With our upcoming implementation of new form settings and types - we can discuss more in-depth how we want to format the form settings object.

kjohnson (approved):
Looks good to me.

PR #81, opened 2022-11-03
Design Preview can now switch templates

kjohnson:
## Description

This PR adds a new `template` form setting with a corresponding `<SelectControl />` which switches the template used in the `<DesignPreview />` canvas.

NOTE: There are some styling issues with the `<select />` element that need to be addressed.

## Visuals

image

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/packages/form-builder/src/common/storage/interface.ts
+++ b/packages/form-builder/src/common/storage/interface.ts
@@ -4,5 +4,5 @@ import {Block} from "../../types/block";
 export interface StorageDriver {
     save({blocks, formSettings}: FormData): Promise<void>,
     load(): FormData,
-    preview(blocks: Block[]): Promise<string>,
+    preview(template: string, blocks: Block[]): Promise<string>,
 }
--- a/packages/form-builder/src/App.scss
+++ b/packages/form-builder/src/App.scss
@@ -135,7 +135,7 @@
  * Fixes an issue with inspector select controls covering selected option.
  */
 .components-select-control__input {
-    padding: 8px;
+    //padding: 8px;
     font-weight: normal;
 }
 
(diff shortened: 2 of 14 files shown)


packages/form-builder/src/common/storage/interface.ts
```diff
@@ -4,5 +4,5 @@ import {Block} from "../../types/block";
 export interface StorageDriver {
     save({blocks, formSettings}: FormData): Promise<void>,
     load(): FormData,
-    preview(blocks: Block[]): Promise<string>,
+    preview(template: string, blocks: Block[]): Promise<string>,
```
kjohnson, line 7: @jonwaldstein I'm considering updating the signature of the `preview` method to match the `FormData` interface used by `save` and `load`. Also, if `blocks` is a property of `formSettings` then the interface can probably be simplified.

jonwaldstein: Ah yeah, that would make sense to do.

jonwaldstein: Also, I think it would make more sense to change `template` to `templateId` just so it's more clear that we are referring to the ID specifically.

jonwaldstein (approved):
@kjohnson great job man!  I did a hot fix for that select padding issue.  Also, I left a reply and  comments on [this](https://github.com/impress-org/givewp-next-gen/pull/81/files/037acf50e675e9064e6c0265589f3ac17d1cdd53#r1013388571) thread you started.  Approving now so those can be addressed now or later.

packages/form-builder/src/common/storage/interface.ts
kjohnson: Resolving via https://github.com/impress-org/givewp-next-gen/issues/82.

PR #91, opened 2022-11-14
Refactor: Design Preview update is now rendered in the background

kjohnson:
## Description

This PR improves the Design Preview by rendering the updated preview in the background and only updating the preview after the document is fully rendered.

## Visuals

Peek 2022-11-14 14-23

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/packages/form-builder/src/components/canvas/DesignPreview.tsx
+++ b/packages/form-builder/src/components/canvas/DesignPreview.tsx
@@ -9,9 +9,9 @@ import DesignPreviewLoading from "@givewp/form-builder/components/canvas/DesignP
 
 const DesignPreview = () => {
     const {blocks, settings: formSettings} = useFormState();
-    const [isLoading, setLoading] = useState(true);
-    const [sourceDocument, setSourceDocument] = useState(null);
-    const [previewHTML, setPreviewHTML] = useState(null);
+    const [isLoading, setLoading] = useState<boolean>(false);
+    const [sourceDocument, setSourceDocument] = useState<string>(null);
+    const [previewHTML, setPreviewHTML] = useState<string>(null);
 
     useEffect(() => {
         setLoading(true);
--- a/packages/form-builder/src/components/canvas/DesignPreview.tsx
+++ b/packages/form-builder/src/components/canvas/DesignPreview.tsx
@@ -1,41 +1,53 @@
 import * as React from 'react';
 import {useEffect, useState} from 'react';
-import DesignPreviewLoading from './DesignPreviewLoading';
 
 import Storage from '@givewp/form-builder/common/storage';
 
 import IframeResizer from 'iframe-resizer-react';
 import {useFormState} from '../../stores/form-state';
+import DesignPreviewLoading from "@givewp/form-builder/components/canvas/DesignPreviewLoading";
 
 const DesignPreview = () => {
     const {blocks, settings: formSettings} = useFormState();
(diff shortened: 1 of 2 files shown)


packages/form-builder/src/components/canvas/DesignPreview.tsx
```diff
@@ -1,41 +1,53 @@
...
+                onLoad={(event) => {
+                    const target = event.target as HTMLIFrameElement;
+                    setPreviewHTML(target.contentWindow.document.documentElement.innerHTML)
+                    setLoading(false)
+                }}
+                srcDoc={sourceDocument}
+                style={{display: 'none'}}
+            />
+        </>
+    )
```
jonwaldstein, line 50: This iframe loading part is not necessary since it will be triggered the same time as useEffect.  All you have to do is check if the `sourceDocument` is not null (and revert removal of `setIsLoading(false)` within useEffect:

```
{isLoading && <DesignPreviewLoading />}
{sourceDocument && <IframeResizer
      srcDoc={sourceDocument}
      checkOrigin={false} /** The srcDoc property is not a URL and requires that the origin check be disabled. */
      style={{
          width: '1px',
          minWidth: '100%',
          border: '0',
      }}
  }
  />}
```

jonwaldstein (changes requested):
Nice update to the UX - just a comment about simplifying the logic.

packages/form-builder/src/components/canvas/DesignPreview.tsx
kjohnson: The `onLoad` even is triggered after the HTML is "ready", which is why I added the event callback. The **rendered** HTML is cached here.

kjohnson:
@jonwaldstein with the feedback resolved is this otherwise approved for merge?

packages/form-builder/src/components/canvas/DesignPreview.tsx
```diff
@@ -1,41 +1,53 @@
...
 import {useFormState} from '../../stores/form-state';
+import DesignPreviewLoading from "@givewp/form-builder/components/canvas/DesignPreviewLoading";
 
 const DesignPreview = () => {
     const {blocks, settings: formSettings} = useFormState();
-    const [sourceDocument, setSourceDocument] = useState('');
-    const [isLoading, setIsLoading] = useState<boolean>();
+    const [isLoading, setLoading] = useState(true);
+    const [sourceDocument, setSourceDocument] = useState(null);
+    const [previewHTML, setPreviewHTML] = useState(null);
```
jonwaldstein, line 15: Let's add the generics to these useStates:

```
const [isLoading, setIsLoading] = useState<boolean>();
const [sourceDocument, setSourceDocument] = useState<string>();
```

As a note I prefer setting default loading states to false because that value should always be programmatically set based on data fetching.

jonwaldstein (changes requested):

packages/form-builder/src/components/canvas/DesignPreview.tsx
kjohnson: Ah, yes - I knew I was forgetting something.

kjohnson: Resolved by https://github.com/impress-org/givewp-next-gen/pull/91/commits/3565f13cfbe381bda8044374cc50ba030a133e8e.

jonwaldstein (approved):
🥇

PR #94, opened 2022-11-17
Feature: Add Design heading, description settings

kjohnson:
## Description

This PR adds the `heading` and `description` settings to the Design tab.

NOTE: This will most likely be refactored over time, but this introduces the settings for use in the rendered form.

## Affects

- Reverts inline label styles for select controls.
- Removes the `styles/_settings.scss` import as unecessary.

## Visuals

image


kjohnson:
@jonwaldstein this has a (seemingly) unrelated failing test related to donation types. I tried to trace the issue, but haven't yet found the issue. Could you take a look?

jonwaldstein:
@kjohnson I love that our tests in here are running properly.  A change in core broke our add-on which makes total sense.  I fixed the issue on develop and updated the tests 👍

packages/form-builder/src/App.tsx
```diff
@@ -22,6 +22,8 @@ const initialState = {
         registration: 'none',
         goalType: 'amount',
         designId: 'classic',
+        heading: __('Support Our Cause'),
+        description: __('Help our organization by donating today! Dnations go to making a difference for our cause.'),
```
jonwaldstein, line 26: It's great we're using the translation function but i'm pretty sure if you don't pass `'give'` as the second argument it wont actually do anything since it needs to locate the textdomain.

jonwaldstein (approved):
I updated the text domains.  everything else looks good 👍

PR #96, opened 2022-11-23
Feature: Add customer feedback popup

kjohnson:
## Description

This PR adds a customer feedback popup/button in the lower right of the form builder.

- The button directs to the feedback URL in an new tab.
- The link in the popup directs to the feedback URL in a new tab and marks the popup as hidden.
- The close icon in the popup marks the popup as hidden.
- Local browser storage is used to keep the popup hidden after use.

## Affects

- Adds bottom padding to the sidebar to avoid obstruction of sidebar settings.

## Visuals

Screenshot of the Feedback Popup in the Form Builder
image

---

Recording of deleting the local browser storage.
Peek 2022-11-23 14-59

---

Recording of updated font size, weight, line height, and alignment.

Peek 2022-11-23 15-01

## Testing Instructions

- Delete the local storage item `givewpNextGenHideFeedback` to restore the popup.


jdghinson (approved):
It looks good!

packages/form-builder/src/feedback/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import Container from './Container'
+import FeedbackButton from "./FeedbackButton";
+
+import {
+    Container as PopupContainer,
+    Header as PopupHeader,
+    Content as PopupContent,
+} from './popup'
+
+const feedbackUrl = 'https://feedback.givewp.com/next-gen';
```
jonwaldstein, line 12: Can we move this url to our `storageData` from the server.  It will just make it easier to update that Url if need be.

packages/form-builder/src/feedback/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return (
+        <Container>
+            {!hidden && (
+                <PopupContainer>
+                    <PopupHeader title={__('Have feedback?', 'givewp')} closeCallback={closeCallback} />
+                    <PopupContent>
+                        {__('Let us know what you think about the form builder to help improve the product experience.', 'givewp')}
+                        {' '}
+                        <a href={feedbackUrl} target="_blank" rel="noopener noreferrer" onClick={closeCallback} style={{color: 'var(--give-primary-600)'}}>
+                            {__('Click here', 'givewp')}
```
jonwaldstein, line 37: The textdomain is `give` not `givewp`.  We'll probably want to update that eventually but lets just make sure to use the right one for now.

jonwaldstein (changes requested):

packages/form-builder/src/feedback/index.tsx
kjohnson: This is a URL that _should not_ change - and if it did it would require a new release, which means a build. I'd rather maintain simplicity here as this component is temporary.

packages/form-builder/src/feedback/index.tsx
kjohnson: Resolved by https://github.com/impress-org/givewp-next-gen/pull/96/commits/0eed4d9f356c9d7ace1a90cca090ba402fbaf163.

jonwaldstein (approved):
🍿

PR #97, opened 2022-11-23
Refactor: register and load form template scripts in cascading order

jonwaldstein:
## Description

This is a big architecture update on how form template related scripts are loaded.  The new loading order looks like this:

**1. Registrar scripts**

This first script loads to assign `window.givewp.form.templates = Object.freeze(defaultFormTemplates)`.  
Our `defaultFormTemplates` is the starting point for add-ons and form designs to interact with.  We make sure to freeze the initial object so that scripts downstream can only modify templates directly to preserve data integrity as much as possible.

**2. Active Form Design**

The active form design (if supplying a JS file) will load _after_ our registrar scripts, so it will have direct access to givewp using the window data:

```
// Get the current version of a template;
const Goal = window.givewp.form.templates.layouts.goal // FC<GoalProps> --> this is being retrieved from core

// Modify the template
window.givewp.form.templates.layouts.goal = MyNewGoal
```

**4. Form Application.**

When the donation form application loads, it simply uses `window.givewp.form.templates` to retrieve the final version of the formTemplates (which is the result of merging core & the active templates - later this will include add-ons).  Throughout the application we are now free to use the templates from the window and grab the value of the template we want to use.

```
const formTemplates = window.givewp.form.templates;

const HeaderTemplate = withTemplateWrapper(formTemplates.layouts.header);
const HeaderTitleTemplate = withTemplateWrapper(formTemplates.layouts.headerTitle);
const HeaderDescriptionTemplate = withTemplateWrapper(formTemplates.layouts.headerDescription);
const GoalTemplate = withTemplateWrapper(formTemplates.layouts.goal);

// render
<HeaderTemplate />
```

**Template Wrapper**

Going down this path allowed me to simplify our rendering process.  Instead of having specific functions for retrieving and rendering specific types of templates, we will now simply grab the templates directly from the active form design and use these template wrappers below.  These will determine how to wrap the template from the template value itself.  The main advantage of this direction is everything becomes self-documenting because the value of the template you are retrieving is typed and used as the generic.

- `withTemplateWrapper`
- `useTemplateWrapper`
- `<TemplateWrapper />`

```
const FieldTemplate = useTemplateWrapper<FieldProps>(formTemplates.fields[node.type]); // React.FC<FieldProps>
const GatewayFieldTemplate = withTemplateWrapper(formTemplates.fields.gateways); // React.FC<GatewayFieldProps>
```

## Affects

There is no visual change but this does affect the whole form rendering process.

## Visuals

There are no visual changes.

## Testing Instructions

Build scripts and make sure the form rendering works and preview work as expected.  There should be no visual differences.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@JasonTheAdams this is ready for review.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/index.tsx
```diff
@@ -1,195 +1,91 @@
...
+ * A hook version of withTemplateWrapper
+ *
+ * @unreleased
+ */
+export function useTemplateWrapper<TemplateProps>(
+    Template: FC<TemplateProps>,
+    htmlTag: keyof JSX.IntrinsicElements = 'div',
+    name?: string
+): FC<TemplateProps> {
+    return useMemo(() => withTemplateWrapper(Template, htmlTag, name), [Template]);
```
JasonTheAdams, line 90: This `useMemo` is not likely to work reliably since the dependency is a function. But maybe it will since the templates don't change once loaded?

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/index.tsx
```diff
@@ -1,195 +1,91 @@
...
-}
-
-export function getHeaderTitleTemplate(): FC<HeaderTitleProps> {
-    return getTemplate('headerTitle', 'layouts');
+/**
+ * This HOC will wrap a template component in our NodeWrapper and automatically figure out what nodeType and type to use as props.
+ *
+ * @unreleased
+ */
+export function withTemplateWrapper<TemplateProps>(
```
JasonTheAdams, line 51: I'm on the fence, but I'm curious what made you decide to keep the template wrappers in the form block code?

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/getFormDesign.ts
```diff
@@ -0,0 +1,10 @@
+import {FormDesign} from '@givewp/forms/types';
+
+/**
+ * Get the active form design
+ *
+ * @unreleased
+ */
+export default function getFormDesign(): FormDesign {
```
JasonTheAdams, line 8: I'm a little nervous about abstracting the long-form path to getting the design. The path itself is critical, and if we abstract it then it won't feel as much to us to be the case. This obviously wouldn't be used outside of core.

src/NextGen/DonationForm/resources/types.ts
```diff
@@ -200,14 +227,26 @@ export function isField(node: Node): node is Field {
     return node.nodeType === 'field';
 }
 
+export function fieldExists(field: Field, template: FormDesign): boolean {
+    return template.fields.hasOwnProperty(field.type);
+}
```
JasonTheAdams, line 232: Couple thoughts:
1. Let's move the `{node}Exists` functions to the utilities (or somewhere not here) as they're not really type functions; they're template functions.
2. Let's rename them to `{node}TemplateExists` to be a bit clearer as to what we're checking.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/registerFieldAndBuildProps.tsx
```diff
@@ -2,23 +2,26 @@ import {Field} from '@givewp/forms/types';
...
 import {UseFormReturn} from 'react-hook-form';
 import buildRegisterValidationOptions from './buildRegisterValidationOptions';
-import {getFieldErrorTemplate, getFieldLabelTemplate} from '../templates';
 import getErrorByFieldName from './getErrorByFieldName';
+import {withTemplateWrapper} from '@givewp/blocks/form/app/templates';
+import getFormDesign from '@givewp/blocks/form/app/utilities/getFormDesign';
+
+const formDesign = getFormDesign();
+const LabelTemplate = withTemplateWrapper(formDesign.layouts.fieldLabel);
+const ErrorMessageTemplate = withTemplateWrapper(formDesign.layouts.fieldError);
```
JasonTheAdams, line 11: Wrapping these feels like markup overkill.

src/NextGen/DonationForm/FormDesigns/DefaultFormDesign/js/layouts/NodeWrapper.tsx
```diff
@@ -0,0 +1,17 @@
...
+import classNames from 'classnames';
+import type {NodeWrapperProps} from '@givewp/forms/propTypes';
+
+export function NodeWrapper({nodeType, type, name, htmlTag: Element = 'div', children}: NodeWrapperProps) {
+    return (
+        <Element
+            className={classNames({
+                [`givewp-${nodeType}`]: nodeType,
+                [`givewp-${nodeType}-${type}`]: type,
+                [`givewp-${nodeType}-${type}-${name}`]: name,
```
JasonTheAdams, line 11: Isn't `name` optional?

src/NextGen/DonationForm/resources/propTypes.ts
```diff
@@ -1,21 +1,117 @@
...
+import {Element, Field, FormDesign, Gateway, Group, Section as SectionType, SelectOption} from '@givewp/forms/types';
+import {FieldErrors, UseFormRegisterReturn} from 'react-hook-form';
+import {FC, FormHTMLAttributes, ReactNode} from 'react';
 
 export interface FieldProps extends Field {
     inputProps: UseFormRegisterReturn;
-    Label: FC;
-    ErrorMessage: FC;
+    Label: FC<FieldLabelProps | {}>;
+    ErrorMessage: FC<FieldErrors | {}>;
```
JasonTheAdams, line 8: Why the optional empty object? Same for the `NameProps` below.

src/NextGen/DonationForm/resources/propTypes.ts
```diff
@@ -1,21 +1,117 @@
...
+    text: string;
+}
+
+export interface HeaderTitleProps {
+    text: string;
+}
+
+export interface NodeWrapperProps {
+    nodeType: keyof FormDesign & string;
+    // TODO: make type work as this keyof (keyof FormDesign) & string
```
JasonTheAdams, line 112: I recommend checking out the new `satisfies` operator in 4.9, in case that helps: https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies

src/NextGen/DonationForm/Registrars/resources/registrars.ts
```diff
@@ -1,54 +1,35 @@
...
+}
+
+window.givewp = {
+    ...window.givewp,
+    gateways: new GatewayRegistrar(),
+    form: {
+        ...window.givewp.form,
+        designs: new FormDesignRegistrar(),
+    },
+};
```
JasonTheAdams, line 35: This could potentially get expensive as more stuff gets shoved in here. Let's avoid rebuilding this object unless we have a clear reason. Just mutate:
```js
window.givewp.gateways = new GatewayRegistrar();
window.givewp.form.designs = new FormDesignRegistrar();
```

JasonTheAdams (changes requested):
Awesome work, @jonwaldstein! There's a lot of really good stuff here, and it's properly cascading, as well as simplifying the application layer. 🙌 

I originally wrote what's below, but I'm really flipping back and forth. You've taken an interesting approach here where _everything_ is a FormDesign, and they just cascade. So the defaults, add-ons, and the "form design" itself are all the same. That even leaves room in the future for parent-child form designs. Read what I wrote, let's chat, and poke it with a stick a bit more.

---

Small naming thing that started coming up in my mind as I was working through this. I'm just processing a bit here: I think of the system of retrieving node components as the Template API. That API has three pieces:
1. Core Templates
2. Add-on Templates
3. Form Design

The purpose of the Template API is to manage the hierarchy of nodes and to determine how each node renders — via a cascading component structure. The Core Templates is a list of default templates that exist out-of-the box. The Add-on templates extend (or add onto) the template system without necessarily caring where the templates come from. Finally, the Form Design is a system which:
- Has a unique name
- Overrides the template hierarchy
- Has its own settings (in the future)

So the Form Design interacts with the Template API, but is distinct from it. So, for example, to get a node template could more accurately be `window.givewp.form.templates.fields.text`.

It would probably be good to properly document this. But I wanted to make sure we're on the same page as not to get into naming confusion as we use and document this. I'm open to discussion on this for sure; the point is to be on the same page.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/registerFieldAndBuildProps.tsx
jonwaldstein: Both of these are currently using the node wrapper from the previous `getTemplate()` function.  I agree it _may_ be overkill but there's also something to say about consistency of the template wrapper being used that we should  discuss.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/index.tsx
jonwaldstein: TBH, I didn't put much thought about _where_ these should live - I was just working from this file as it was previously.  I will say the template wrapper is very specific to the form rendering process and needs to work with the final form design template.  This is not something we want to expose, so I think it can either stay here or move to the root of the domain.

src/NextGen/DonationForm/resources/propTypes.ts
jonwaldstein: This lets us render the components (when props are already applied) in our templates, yet still have the props available:

Screen Shot 2022-11-29 at 10 55 34 PM

src/NextGen/DonationForm/FormDesigns/DefaultFormDesign/js/layouts/NodeWrapper.tsx
jonwaldstein: yes, this is a feature of the classNames library where it only adds the className if the argument is not null.

src/NextGen/DonationForm/Registrars/resources/registrars.ts
jonwaldstein: resolved d005c6ac

src/NextGen/DonationForm/resources/types.ts
jonwaldstein: resolved e971c1f1

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/templates/index.tsx
jonwaldstein: This made me realize that the entire _form block app_ has no more business being inside the block domain anymore - since all the block does is render an iframe now.  I moved the donation form app inside our root domain's resources folder e125aa47 which further isolates the donation form as a standalone application that does not depend on the block itself.

jonwaldstein:
> Awesome work, @jonwaldstein! There's a lot of really good stuff here, and it's properly cascading, as well as simplifying the application layer. 🙌
> 
> I originally wrote what's below, but I'm really flipping back and forth. You've taken an interesting approach here where _everything_ is a FormDesign, and they just cascade. So the defaults, add-ons, and the "form design" itself are all the same. That even leaves room in the future for parent-child form designs. Read what I wrote, let's chat, and poke it with a stick a bit more.
> 
> Small naming thing that started coming up in my mind as I was working through this. I'm just processing a bit here: I think of the system of retrieving node components as the Template API. That API has three pieces:
> 
> 1. Core Templates
> 2. Add-on Templates
> 3. Form Design
> 
> The purpose of the Template API is to manage the hierarchy of nodes and to determine how each node renders — via a cascading component structure. The Core Templates is a list of default templates that exist out-of-the box. The Add-on templates extend (or add onto) the template system without necessarily caring where the templates come from. Finally, the Form Design is a system which:
> 
> * Has a unique name
> * Overrides the template hierarchy
> * Has its own settings (in the future)
> 
> So the Form Design interacts with the Template API, but is distinct from it. So, for example, to get a node template could more accurately be `window.givewp.form.templates.fields.text`.
> 
> It would probably be good to properly document this. But I wanted to make sure we're on the same page as not to get into naming confusion as we use and document this. I'm open to discussion on this for sure; the point is to be on the same page.

I'll leave my thoughts here and we can chat through these concepts together to determine a concluded direction.

In my mind, the Template API can be thought of as a simple config that holds the values the form will use to render.

This is applied through a cascading architecture of _Form Template Schemas_ that result into a final _Form Design_ that the form can understand and retrieve values from.

Here is what that practically looks like using some different naming conventions:

1. GiveWP Core loads the initial Form Template Schema
2. Add-ons extend the Form Template Schema
3. Form designs override and extend the Form Template Schema
4. Form application converts the cascaded Form Template Schemas into a Form Design.
5. Form application uses values from the Form Design.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/getFormDesign.ts
JasonTheAdams: My point here is to keep the number of ways of accessing node templates reduced to have a consistent (and shared with 3rd parties) DX.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/app/utilities/registerFieldAndBuildProps.tsx
JasonTheAdams: Mm... For these very specific layouts it just feels like markup cruft. I appreciate the consistency when it comes to fields, elements, and such, but I don't want to follow the convention for its own sake. I'm more interested in what the resulting markup is, and in this case it's too much.

src/NextGen/DonationForm/resources/types.ts
```diff
@@ -82,29 +100,43 @@ export interface Gateway {
...
-    };
+export interface FormTemplateFields {
+    amount: FC<AmountProps>;
+    text: FC<FieldProps>;
+    textarea: FC<FieldProps>;
+    email: FC<FieldProps>;
+    hidden: FC<FieldProps>;
+    gateways: FC<GatewayFieldProps>;
+    select: FC<SelectFieldProps>;
+}
```
JasonTheAdams, line 111: Let's import the `defautlFormTemplates` and use that to provide the specific types, then add a dynamic type for the ones added later. Here's an [example of what I mean](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwGIGEC+cBmUIjgIigFMBDAYxnwFgAoWgEyLIBsTi4yIA7AZ3gBUiADxgAuOKgA8JLgE8AfAyat2nXgOEwAgsRLipMhbVpq+cRphIBXZjCTQQg8KxhEecALyJacX1mBEzPQ84gg+fhGuIuKCIgA04RG+UTBspDGaOqSJcGi0eXQ0MLJgRBIOTmAuREgBQe5exaUQmOZEljZ2FUTOJK48AHSYdcFwAGRwAEpM0PSSfFDAXADmcRIo0nLyioWMLGkc3GbTZFZQiyv6G4Y7Jkfww4HB+t29rrVPDd40kZoZ8TkUmk9HBYtpdAkfn5iKdzktluITmcLst8kA). The same can be done for elements and groups.

As a note, the `import type` declarative can be used on things not just types (in this case defaultFormTemplates). It just tells it not to actually import it when bundling.

JasonTheAdams (changes requested):
Loving where this is going, @jonwaldstein! I think we're almost there. I left a couple notes.

I am a bit confused since it looks like, in `api.ts` the extend methods and such are how nodes are added, but then in the Classic template it looks like it's just mutating the object. Am I missing something?

jonwaldstein:
> Loving where this is going, @jonwaldstein! I think we're almost there. I left a couple notes.
> 
> I am a bit confused since it looks like, in `api.ts` the extend methods and such are how nodes are added, but then in the Classic template it looks like it's just mutating the object. Am I missing something?

Yeah so that api file was just leftover from when I had the api endpoint.  I decided to go simple by freezing the `defaultFormTemplates` and allow for mutating the template nodes directly.  Removed that api file 39216a06

src/NextGen/DonationForm/resources/types.ts
jonwaldstein: I see what your saying, just let the `defaultFormTemplate` dictate what the FormTemplates type is.  I decided to simplify this whole thing here b7c7a4d3 because defining the types is kind of pointless now, since the only place it was being used was for the initial object.

I'm actually already using the typeof the templates from the window for the template wrapper too so everything just stems from that initial `defaultFormTemplate` which makes things easier.

src/NextGen/DonationForm/resources/registrars/templates/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import SelectField from './fields/Select';
+import Gateways from './fields/Gateways';
+import Paragraph from './elements/Paragraph';
+import FieldLabel from './layouts/FieldLabel';
+import FieldError from './layouts/FieldError';
+import Header from './layouts/Header';
+import HeaderTitle from './layouts/HeaderTitle';
+import HeaderDescription from './layouts/HeaderDescription';
+import Goal from './layouts/Goal';
+import {NodeWrapper} from './layouts/NodeWrapper';
```
JasonTheAdams, line 20: @jonwaldstein Let's make this export as a default to be consistent with the others.

JasonTheAdams (approved):
This is awesome, @jonwaldstein! Not only is the system simpler, more dynamic, and more reliably cascaded, but we did it in a way where there's even less code. That's a success!

I left one super minor suggestion, but I'm going to approve this. I do wonder if we'll run into any type issues when adding nodes in add-ons since the default type doesn't allow for additional properties, but we can easily tweak that at that time. This is fundamentally complete. 💪

src/NextGen/DonationForm/resources/registrars/templates/index.ts
jonwaldstein: resolved fec9497a

PR #99, opened 2022-12-02
Feature: Add custom field support in Donation Details, Donor Details

kjohnson:
## Description

This PR adds support for displaying custom fields/values in the Donation Details and Donor Details screens in the WordPress/GiveWP admin dashboard.

## Affects

- Adds a "Display Settings" section to custom fields.
- Adds the "Display in Admin" toggle in the Display Settings section.
- Adds Controller/View for the Donation Details metabox.
- Adds Controller/View for the Donor Details table.
- Adds `shouldDisplayInAdmin` macro to `Give\Framework\FieldsAPI\Field`.

### Display Settings inspector section
Display Settings inspector section

### Donation Details metabox
Donation Details metabox

### Donor Details table
Donor Details table

## Testing Instructions

- Add a custom field
- Configure to "Display in Admin"
- Submit the donation form.
- Custom field label/value should be displayed in the Donation Details screen.
- Repeat (using "Store as Donor Meta" setting) for Donor custom field.

### DIFF
--- a/packages/form-builder/src/blocks/fields/field.js
+++ b/packages/form-builder/src/blocks/fields/field.js
@@ -3,13 +3,13 @@ import {__} from '@wordpress/i18n';
 import defaultSettings from './settings';
 import DefaultFieldSettings from './settings/DefaultFieldSettings';
 import {useFieldNameValidator} from '../../hooks';
-import {InspectorAdvancedControls} from '@wordpress/block-editor';
-import {ExternalLink, PanelRow, TextControl, ToggleControl} from '@wordpress/components';
+import {InspectorAdvancedControls, InspectorControls} from '@wordpress/block-editor';
+import {ExternalLink, PanelBody, PanelRow, TextControl, ToggleControl} from '@wordpress/components';
 import slugify from '../../common/slugify';
 import {useCallback} from '@wordpress/element';
 
 function FieldSettings({attributes, setAttributes}) {
-    const {fieldName, label, storeAsDonorMeta} = attributes;
+    const {fieldName, label, storeAsDonorMeta, displayInAdmin, displayInReceipt} = attributes;
     const validateFieldName = useFieldNameValidator();
 
     const updateFieldName = useCallback(
@@ -46,7 +46,33 @@ function FieldSettings({attributes, setAttributes}) {
                     }
                 }}
             />
+            <InspectorControls>
+                <PanelBody title={__('Display Settings', 'give')} initialOpen={true}>
+                    <PanelRow>
+                        <ToggleControl
+                            label={__('Display in Admin', 'give')}
+                            checked={displayInAdmin}
+                            onChange={() => setAttributes({displayInAdmin: !displayInAdmin})}
+                        />
+                    </PanelRow>
+                    <PanelRow>
+                        <ToggleControl
+                            label={__('Display in Receipt', 'give')}
+                            checked={displayInReceipt}
+                            onChange={() => setAttributes({displayInReceipt: !displayInReceipt})}
+                        />
+                    </PanelRow>
+                </PanelBody>
+            </InspectorControls>
             <InspectorAdvancedControls>
+                <PanelRow>
+                    <ToggleControl
+                        label={__('Store as Donor Meta', 'give')}
+                        checked={storeAsDonorMeta}
+                        onChange={() => setAttributes({storeAsDonorMeta: !storeAsDonorMeta})}
+                        help={__('By default, fields are stored as Donation Meta', 'give')}
+                    />
+                </PanelRow>
                 <PanelRow>
                     <TextControl
                         label={__('Field Name', 'give')}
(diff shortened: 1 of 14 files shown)


src/NextGen/DonationForm/DataTransferObjects/DonationFormGoalData.php
```diff
@@ -80,7 +80,7 @@ public function toArray(): array
             'currentAmount' => $currentAmount,
             'targetAmount' => $this->targetAmount,
             'label' => $this->goalType->isDonors() ? __('donors', 'give') : __('donations', 'give'),
-            'progressPercentage' => !$currentAmount ? 0 : ($currentAmount / $this->targetAmount) * 100
+            'progressPercentage' => !$currentAmount || !$this->targetAmount ? 0 : ($currentAmount / $this->targetAmount) * 100
```
kjohnson, line 83: This prevents a division by zero error when the `targetAmount` is not set.

src/NextGen/CustomFields/Controllers/DonorDetailsController.php
```diff
@@ -0,0 +1,68 @@
...
+use Give\Donations\Models\Donation;
+use Give\Donors\Models\Donor;
+use Give\NextGen\CustomFields\Views\DonorDetailsView;
+use Give\NextGen\DonationForm\Models\DonationForm;
+use Give_Donor as LegacyDonor;
+
+/**
+ * @unreleased
+ */
+class DonorDetailsController
```
jonwaldstein, line 14: I noticed these details classes are being called "controllers" but they just have a single invokable method - which I think lends itself to be classified as an action.  Do you foresee adding crud operations to these in the future? If that was the case I could see these `_invoke` methods being called `show`. Otherwise I think some more descriptive actions would work better like `ShowCustomFieldsInDonorDetails`

src/NextGen/CustomFields/ServiceProvider.php
```diff
@@ -0,0 +1,38 @@
...
+ * @unreleased
+ */
+class ServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function register()
+    {
+        Field::macro('shouldDisplayInAdmin', function() {
```
jonwaldstein, line 21: Woah, that's cool!

packages/form-builder/src/blocks/fields/field.js
```diff
@@ -92,6 +110,11 @@ const field = {
                 source: 'attribute',
                 default: false,
             },
+            displayInAdmin: {
+                type: 'boolean',
+                source: 'attribute',
+                default: false,
```
jonwaldstein, line 116: I'm thinking custom fields should be `true` by default to display in admin & receipt.  It's something I see being easily missed by the admin and then immediately confused by lol

jonwaldstein (changes requested):
This is awesome! It's working well, just some minor suggestions.

src/NextGen/CustomFields/Controllers/DonorDetailsController.php
kjohnson: These are kinda like single method invokable controller action things. So I grabbed a couple of those words by random 😅 For real though, I went back and fourth a lot and went with `Controller`. I like your idea of changing `__invoke` to `show` to be more controller like - I think that makes sense here.

src/NextGen/CustomFields/ServiceProvider.php
kjohnson: Right? Macros are great here for extending the Fields API.

src/NextGen/CustomFields/Controllers/DonorDetailsController.php
jonwaldstein: lol I can relate, naming things is hard!

packages/form-builder/src/blocks/fields/field.js
kjohnson: I went with `false` as the default to [maintain parity with the Fields API](https://github.com/impress-org/givewp/pull/5886) and am open to discussing/considering further.

src/NextGen/CustomFields/Controllers/DonorDetailsController.php
kjohnson: Resolved by https://github.com/impress-org/givewp-next-gen/pull/99/commits/ba0ab4964789ad831fb31568a9e6f99e73b6e18e.

src/NextGen/CustomFields/ServiceProvider.php
JasonTheAdams: So... I do love macros, but why not just bake this right into Fields? 😅

kjohnson: That would require a change to Give core, which we can do, but I wanted to start here.

packages/form-builder/src/blocks/fields/field.js
kjohnson: Having slept on it, I agree that `true` would be a better default for this use case.

Resolved, see https://github.com/impress-org/givewp-next-gen/pull/99/commits/2badd835bdfd5df0a772112873e631a63c2c6dc6.

jonwaldstein (approved):
Thanks for the updates, this is great!

PR #102, opened 2022-12-22
Fix: Set consistent border below tab group

kjohnson:
## Description

This PR adds a top border to the `block-card` element, when the Block tab is selected, so that the border separator is consistent with the other tabs.

## Visuals

| Form Tab | Block Tab (before) | Block Tab (after) |
|---|---|---|
|image|image|image|

### DIFF
--- a/packages/form-builder/src/App.scss
+++ b/packages/form-builder/src/App.scss
@@ -349,11 +349,15 @@ svg {
 /** ------------------------------- */
 
 /*
-Add top border to block card (block tab) to match panel body elements.
+Add border bottom to tabs, while removing body top border.
 This creates a consistent separator between the tabs and the rest of the sidebar.
 */
-.block-editor-block-card {
-    border-top: 1px solid #e0e0e0;
+.components-tab-panel__tabs {
+    border-bottom: 1px solid #e0e0e0;
+}
+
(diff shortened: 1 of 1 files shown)


packages/form-builder/src/App.scss
```diff
@@ -347,3 +347,13 @@ svg {
...
+
+/*
+Add top border to block card (block tab) to match panel body elements.
+This creates a consistent separator between the tabs and the rest of the sidebar.
+*/
+.block-editor-block-card {
+    border-top: 1px solid #e0e0e0;
+}
+
+/** ------------------------------- */
```
jonwaldstein, line 359: @kjohnson I know this is super small but I noticed it's technically possible to select the block tab without a block selected which looks like this:

Screen Shot 2022-12-22 at 4 26 36 PM


So let's target the tabs themselves so even in those weird cases the border is still there lol


```suggestion

/*
Add border bottom to tabs, while removing body top border.
This creates a consistent separator between the tabs and the rest of the sidebar.
*/
.components-tab-panel__tabs {
    border-bottom: 1px solid #e0e0e0;
}

.components-panel__body {
    border-top: none;
}

/** ------------------------------- */
```

PR #104, opened 2022-12-22
Refactor: Remove layout margin when not selected

kjohnson:
## Description

This PR removes the extra bottom margin from the section block when not selected. Previously, this bottom margin was added as a placeholder to prevent shifting when the section block was selected and the block appender was rendered in that space.

## Visuals

| Before | After |
| --- | --- |
| Screenshot from 2022-12-22 11-43-23 | Screenshot from 2022-12-22 11-44-17  |

---
  - https://app.asana.com/0/0/1203547864809100

### DIFF
--- a/packages/form-builder/src/App.scss
+++ b/packages/form-builder/src/App.scss
@@ -168,10 +168,6 @@ svg {
         gap: 24px;
         flex-direction: column;
     }
-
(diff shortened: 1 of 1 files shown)


jonwaldstein (approved):
I modified this so the button appears within the padding bottom and never adds margin to prevent height jumping.

PR #113, opened 2023-01-11
Feature: add stateless donation confirmation receipt routing, template, and implementation

jonwaldstein:
## Description

This introduces stateless receipts.  This means we will not be relying on the session to store information about our donation to display a confirmation receipt. 

Here are the most common scenarios for receipts:

1. Onsite gateways will redirect to the receipt within their iframe.
2. Offsite gateways will redirect back to the origin page of the donation form - which will load the receipt where the donation form was.

**Donation Confirmation Receipt View Route**

This url will mainly be used _under-the-hood_ to display the confirmation receipt in various scenarios.  Typically this means this will be the url of an iframe.  You can also use this route to view the receipt as a standalone page with nothing else on it.

```
$receiptId = $donation->purchaseKey

/?givewp-route=donation-confirmation-receipt-view&receipt-id={receiptId}
```

**Donation Confirmation Receipt Url**

This is for displaying the receipt in place of an embedded donation form.

```
{originUrl}/?givewp-event=donation-completed&givewp-listener=show-donation-confirmation-receipt&givewp-receipt-id={receiptId}&givewp-embed-id={embedId}
```

**Donation Flow & Controller Updates**

In order to preserve our framework's handler for `PaymentComplete` we are now filtering the default `give_get_success_page_uri` with our receipt view url.  This is primary used for onsite gateways.

Offsite gateways are more complicated but will now have an easy way to grab the redirectReturnUrl straight from `$gatewayData['redirectReturnUrl']`.  Once we start implementing _existing_ gateways into Next Gen we will filter this url on the legacy adapter so the gateways will always receive a value - but it will be determined by the request.

## Affects

This touches the next gen donation form flow including the current gateways and a new offsite test gateway.

## Visuals

Screen Shot 2023-01-13 at 10 06 20 AM

## Testing Instructions

- Rebuild assets
- Add a new donation form block
- Enable Next Gen Gateways (_Next Gen Test Gateway, Next Gen Test Gateway Offsite, and Next Gen Stripe_)
- Donate using all gateways
- The form should be directed to new receipt after completion
- Now you can try adding custom fields to form to see them on receipt
- You can also add additional form blocks to the page - the receipt will only display in place of the one completed
- You can also visit the standalone receipt view url for any donation using the purchase ke which can be found in the "Donation Meta" box (labeled "Key") in the single donation admin. `/?givewp-route=donation-confirmation-receipt-view&receipt-id={key}`

## Pre-review Checklist

-   [x] Includes unit tests


jonwaldstein:
@kjohnson this is ready for review 😅

src/NextGen/DonationForm/Blocks/DonationFormBlock/Controllers/BlockRenderController.php
```diff
@@ -44,12 +47,35 @@ public function render(array $attributes)
...
+    {
+        $routeListener = new RouteListener(
+            'donation-completed',
+            'show-donation-confirmation-receipt'
+        );
+
+        return $routeListener->isValid($_GET, function ($request) use ($embedId) {
+            $isset = isset($request['givewp-embed-id'], $request['givewp-receipt-id']);
+            
+            return $isset && $request['givewp-embed-id'] === $embedId && strlen($request['givewp-receipt-id']) === 32;
```
kjohnson, line 78: Is there a way around using a magic number here? Can we set that to an Enum or a class constant?

src/NextGen/DonationForm/Blocks/DonationFormBlock/Controllers/BlockRenderController.php
```diff
@@ -44,12 +47,35 @@ public function render(array $attributes)
         $donationForm = DonationForm::find($blockAttributes->formId);
 
         $viewUrl = (new GenerateDonationFormViewRouteUrl())($donationForm->id);
+        $embedId = $blockAttributes->blockId;
+
+        if ($this->shouldDisplayDonationConfirmationReceipt($embedId)) {
+            $receiptId = give_clean($_GET['givewp-receipt-id']);
+
+            $viewUrl = (new GenerateDonationConfirmationReceiptViewRouteUrl())($receiptId);
```
kjohnson, line 55: Instead of mutating the value here, should we use an `if/then/else` to set the variable value?

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/editor/Edit.tsx
```diff
@@ -13,10 +13,16 @@ import {useCallback} from 'react';
...
  */
-export default function Edit({attributes, setAttributes}: BlockEditProps<any>) {
-    const {formId} = attributes;
+export default function Edit({clientId, attributes, setAttributes}: BlockEditProps<any>) {
+    const {formId, blockId} = attributes;
     const {formOptions, isResolving} = useFormOptions();
 
+    useEffect(() => {
+        if (!blockId) {
+            setAttributes({blockId: clientId});
```
kjohnson, line 22: Have we looked into the possibility of conflicts with re-using the `clientId` as a unique identifier outside of the block editor?

The `clientId` is used to track the block within the editor and is not persisted in the post content. This value changes each time the post editor is loaded, so the uniqueness of the value is limited to editor session.

src/NextGen/DonationForm/Controllers/DonateController.php
```diff
@@ -40,8 +39,9 @@ public function donate(DonateControllerData $formData, PaymentGateway $registere
 
         $this->saveCustomFields($form, $donation, $formData->getCustomFields());
 
-        // setting sessions is required for legacy receipts
-        $this->setSession($donation, $donor);
+        $this->filterSuccessPageUri($formData->getDonationConfirmationReceiptViewRouteUrl($donation));
```
kjohnson, line 42: I would consider a different name here, since in the context of WordPress "filter" insinuates a return value that may or may not be the same as the argument.

src/NextGen/DonationForm/Controllers/DonateController.php
```diff
@@ -92,37 +92,52 @@ private function getOrCreateDonor(
...
+    /**
+     * Use our new receipt url for the success page uri
+     *
+     * @unreleased
+     *
+     * @return void
+     */
+    protected function filterSuccessPageUri(string $filteredUrl)
+    {
+        add_filter('give_get_success_page_uri', static function ($url) use ($filteredUrl) {
```
kjohnson, line 113: In general, should we be adding filters like this inside of the controller? Usually filters are added to a Service Provider. This would be a big break in convention if we start including filters within controllers.

kjohnson: I see why you put the filter in the controller, but I think it warrants a discussion.

src/NextGen/DonationForm/DataTransferObjects/DonationConfirmationReceiptViewRouteData.php
```diff
@@ -0,0 +1,39 @@
...
+
+        return $self;
+    }
+
+    /**
+     * @unreleased
+     */
+    protected function isValidReceiptId(string $receiptId): bool
+    {
+        return strlen($receiptId) === 32;
```
kjohnson, line 37: Here is another instance of that magic number `32`. It would probably be helpful to centralize this value with documentation.

src/NextGen/DonationForm/ViewModels/DonationConfirmationReceiptViewModel.php
```diff
@@ -0,0 +1,214 @@
...
+    {
+        $receipt = new DonationReceipt($this->donation);
+        $receipt->settings->addSetting(
+            'heading',
+            __("Hey {$this->donation->firstName}, thanks for your donation!", 'give')
+        );
+        $receipt->settings->addSetting(
+            'description',
+            __(
+                "{$this->donation->firstName}, your contribution means a lot and will be put to good use in making a difference. We’ve sent your donation receipt to {$this->donation->email}.",
```
kjohnson, line 53: Are these string values customizable somewhere or are we using these hard-coded strings?

src/NextGen/DonationForm/ViewModels/DonationFormViewModel.php
```diff
@@ -117,7 +117,9 @@ public function exports(): array
 
         return [
             'donateUrl' => $donateUrl,
-            'successUrl' => give_get_success_page_uri(),
+            'inlineRedirectRoutes' => [
```
kjohnson, line 120: Where is this being used? Also, it seems odd to be exporting a hard-coded value here.

kjohnson: I see the reference in `app/form/Form.tsx`, but what is the reason for localizing this hard-coded value?

src/NextGen/DonationForm/Controllers/DonateController.php
jonwaldstein: Yes I realize this is not ideal, It's tricky because we need to completely replace that value based on the current request - that would not be available in a service provider.  I should note that this is not intended to be for the long term, eventually we will want to update the gateway api to handle redirecting to the appropriate success page.

src/NextGen/DonationForm/Blocks/DonationFormBlock/Controllers/BlockRenderController.php
jonwaldstein: Sure, the number 32 is just the standard character length of an md5 hash.  I can add something that makes that more clear.

src/NextGen/DonationForm/Controllers/DonateController.php
JasonTheAdams: I agree it's not ideal, but so long as the API for the gateways themselves is declarative and immutable, stuff like this doesn't bother me too much. Make sure to leave a note here that as to the fact that it's temporary and how it ideally ought to be handled in the future.

src/NextGen/DonationForm/Blocks/DonationFormBlock/resources/editor/Edit.tsx
jonwaldstein: I'm only using it as an additional data attribute on the iframe being rendered to target our form for displaying a receipt.  They only have to be unique for our blocks specifically, so i'm not worried about conflicts.

src/NextGen/DonationForm/ViewModels/DonationConfirmationReceiptViewModel.php
jonwaldstein: This is a parity setting that will come from the form builder receipt settings.

src/NextGen/DonationForm/ViewModels/DonationFormViewModel.php
jonwaldstein: Oh yeah, my thought was there will be be more inline routes added here that will come from some kind of registrar.  Another example will be a form error route.  I want the front-end to be aware that these routes exist but have the backend supply them.  The registrar will likely tie into the the new Route facade.

JasonTheAdams:
@jonwaldstein Just a heads up I was testing this for a demo and noticed the Company field doesn't show up in the receipt.

src/NextGen/DonationForm/Blocks/DonationFormBlock/Controllers/BlockRenderController.php
jonwaldstein: resolved here for now [fd7d6bc](https://github.com/impress-org/givewp-next-gen/pull/113/commits/fd7d6bc273399386061b1aecced148c47e7fe357).

src/NextGen/DonationForm/DataTransferObjects/DonationConfirmationReceiptViewRouteData.php
jonwaldstein: Resolved by adding description for now. fd7d6bc

jonwaldstein:
> @jonwaldstein Just a heads up I was testing this for a demo and noticed the Company field doesn't show up in the receipt.

Good catch, resolved af934f21

src/NextGen/DonationForm/Blocks/DonationFormBlock/Controllers/BlockRenderController.php
jonwaldstein: resolved 86a40b2d

src/NextGen/DonationForm/Controllers/DonateController.php
jonwaldstein: resolved b55ef555

jonwaldstein:
@kjohnson feedback addressed and tests are passing.

PR #114, opened 2023-01-17
Feature: Add Form Page for Next Gen Donation Forms

kjohnson:
## Description

This PR adds Form Pages for Next Gen Donation Forms, including a unique page slug (`post_name`) and dynamically rendered page content using the Next Gen Form Block.

## Affects

- Updates the `Storage.save()` signature to return `FormSettings`. This is used to relay update the page slug (from the server).

## Visuals

Screenshot 2023-01-20 at 11 20 41 AM


packages/form-builder/src/App.tsx
```diff
@@ -33,6 +33,7 @@ const initialState = {
         ),
         primaryColor: '#69b86b',
         secondaryColor: '#f49420',
+        pageSlug: 'new-form',
```
kjohnson, line 36: This is really just a placeholder. The default value is generated by the server for uniqueness when the form is created (which is before the builder loads for the first time).

packages/form-builder/src/containers/HeaderContainer.jsx
```diff
@@ -24,7 +24,8 @@ const HeaderContainer = ({
         setSaving(true);
         Storage.save({blocks, formSettings})
             .catch((error) => alert(error.message))
-            .finally(() => {
+            .then((updatedSettings) => {
+                dispatch(setFormSettings({pageSlug: updatedSettings.pageSlug}));
```
kjohnson, line 28: @jonwaldstein This is where we are updating the page slug in the builder, as it may have been updated on the server (for uniqueness).

src/FormBuilder/ServiceProvider.php
```diff
@@ -46,5 +46,12 @@ public function boot()
                 'newFormUrl' => FormBuilderRouteBuilder::makeCreateFormRoute()->getUrl(),
             ]);
         });
+        
+        /** Dynamically renders Next Gen Forms using the Next Gen Form block. */
+        add_action('give_before_single_form', function() {
```
kjohnson, line 51: This is a work in progress. I ran into some issue is replacing the form content and will need a separate PR to really address how this should work. But I wanted to take a step in that direction and not further convolute this change.

src/FormBuilder/ViewModels/FormBuilderViewModel.php
```diff
@@ -19,6 +19,7 @@ public function storageData(int $donationFormId): array
         $donationForm = DonationForm::find($donationFormId);
 
         return [
+            'formPageURL' => add_query_arg(['p' => $donationFormId], site_url()),
```
kjohnson, line 22: This uses the `?p={id}` query parameter so that it always works, regardless of the page slug setting (which may not be unique when changed, that is, until the server can update it).

jonwaldstein:
@kjohnson 

I'm getting both a next gen and legacy form rendered on my form page:


Screenshot 2023-01-19 at 4 23 57 PM

kjohnson:
> I'm getting both a next gen and legacy form rendered on my form page

Yeah, I mentioned that previously on our calls. There is currently a limitation on how the Give forms are rendered. This is a temporary solution and I have ideas as to how to resolve the underlying issue - but, again, don't want to convolute this pull request.

src/FormBuilder/ServiceProvider.php
jonwaldstein: We're going to need to override the single give form completely.  Not sure if thats possible via a plugin but maybe we can do `template_include` to replace that file with ours.

jonwaldstein (changes requested):

jonwaldstein:
> > I'm getting both a next gen and legacy form rendered on my form page
> 
> Yeah, I mentioned that previously on our calls. There is currently a limitation on how the Give forms are rendered. This is a temporary solution and I have ideas as to how to resolve the underlying issue - but, again, don't want to convolute this pull request.

I think it would make the most sense to resolve this in this PR since it's purpose is to add a form page.  I commented on that part with an idea.

packages/form-builder/src/settings/form-summary/page-slug.js
```diff
@@ -0,0 +1,49 @@
...
+                <TextControl
+                    label={'Permalink'}
+                    value={pageSlug}
+                    onChange={(pageSlug) => dispatch(setFormSettings({pageSlug}))}
+                    help={'The last part of the URL.'}
+                />
+                <div>View Page</div>
+                <ExternalLink href={window.storageData.formPageURL}>{'/donations/' + pageSlug}</ExternalLink>
+            </div>
+        ) }
```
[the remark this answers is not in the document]
kjohnson: Resolve as a part of https://github.com/impress-org/givewp-next-gen/pull/114/commits/731ad31e35465c8ad32ced9fb2b075fecf5072d2. I localized the post type's rewrite slug instead of the full URL in order to save space in the UI.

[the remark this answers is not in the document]
jonwaldstein: Nice!

src/FormBuilder/ServiceProvider.php
jonwaldstein: I just tested this out and it works:

```
 add_filter('template_include', static function ($template) {
      if (is_single() && get_post_type() === 'give_forms') {
          $template = GIVE_NEXT_GEN_DIR . 'templates/givewp-single-form.php';
      }

      return $template;
  }, 99);
```

jonwaldstein (approved):
@kjohnson great job!

PR #121, opened 2023-01-24
Fix: Add back color setting defaults

kjohnson:
## Description

This PR restores the default color options in the color dropdown. The default values were removed by Gutenberg to be more dynamic.

Also, this affixes the color dropdown to the popover slot so that it does not scroll with the form.

## Visuals

image

## Testing Instructions

Run `npm run build:form-builder` and check that the color settings in Design Mode have default options to choose from. Also, the color picker should not scroll with the form, but rather scroll with the sidebar.


jonwaldstein (approved):
nice!

PR #122, opened 2023-01-25
Fix: Stripe gateway customer retrieval

jonwaldstein:
## Description

This PR fixes an issue with a donor making a donation through different stripe accounts.  In order to achieve this we needed to catch Stripe's `InvalidRequestException` exception when a customer is not found and then create a new customer.  This is actually how the logic works in legacy.  We also are checking if a customer has the is_deleted flag so we can re-create the customer.

**Note**
While this logic has been working for us for some time, I am seeing room for improvement.  

**Multiple Stipe account**
GiveWP has the ability to use different Stripe accounts.  From what I can tell, our donor meta does not account for this - meaning anytime a donor uses stripe for a donation, the customer ID for _that_ account is saved to their meta - overwriting the previous value.  We should really have an identifier for that customer ID that connects back to the specific stripe account it's for.  That could be something like a key / value pair with the stripe account as the key and customer ID as the value:

```
_give_stripe_customer_id{_test} = [
'acc_12345' => 'cus_12345',
'acc_54321' => 'cus_54321'
];
```

**Stripe API & searching**
Newer versions of the Stripe api feature a search functionality.  This would be a much better use case for us when try to retrieve a customer record.  As pointed out earlier, the only way to know if a customer _does not_ exist is by catching a _resource_missing_ exception.  Searching for a customer based on ID and/or email appears to be a more appropriate way of looking up a customer.

```
$stripe->customers->search(['query' => 'email:\'sally@rocketrides.io\'']);
```

## Affects

- Next Gen Stripe gateway

## Visuals

N/A

## Testing Instructions

- Donate with a Stripe account
- Now connect to a different Stripe account (set as default)
- Donate with the same email as before
- The donation should go through properly

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
+++ b/src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
@@ -6,6 +6,7 @@ use Give\Donations\Models\Donation;
 use Give\Donations\Models\DonationNote;
 use Give\Framework\Exceptions\Primitives\Exception;
 use Give\Framework\Support\ValueObjects\Money;
+use Give\Log\Log;
 use Give\PaymentGateways\Exceptions\InvalidPropertyName;
 use Give\PaymentGateways\Gateways\Stripe\Actions\SaveDonationSummary;
 use Give\PaymentGateways\Stripe\ApplicationFee;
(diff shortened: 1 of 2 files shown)


jonwaldstein:

kjohnson:
@jonwaldstein I think we should capture these improvements somewhere, like a `@todo` or a new issue, so that they don't get lost in the merged/closed PR.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
```diff
@@ -34,40 +36,52 @@ protected function generateStripePaymentIntent($accountId, Money $amount): Payme
...
+        } else {
+            try {
+                // if the donor has a stripe ID, try retrieving the customer from Stripe.
+                $customer = $this->retrieveCustomer($donorCustomerId, $connectAccountId);
+
+                // if the customer is deleted, create a new customer
+                if ($customer->isDeleted()) {
+                    $customer = $this->createCustomer($donation, $connectAccountId);
+                }
+            } catch (InvalidRequestException $e) {
```
kjohnson, line 61: Is there a more specific Exception that we could catch? Nesting an `if` inside of a `catch` can suppress instances of the Exception if not handled by and `if` condition.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
```diff
@@ -34,40 +36,52 @@ protected function generateStripePaymentIntent($accountId, Money $amount): Payme
...
      * @throws Exception
+     * @throws ApiErrorException
      */
     public function getOrCreateStripeCustomerFromDonation(
         string $connectAccountId,
         Donation $donation
-    ): Customer {
+    ): Customer
+    {
         $donorCustomerId = give_stripe_get_customer_id($donation->email) ?? '';
```
kjohnson, line 47: From what I can tell the `?? ''` is redundant. `give_stripe_get_customer_id()` uses an empty string as a default.

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
jonwaldstein: Resolved bb4817dc

src/NextGen/Gateways/Stripe/NextGenStripeGateway/NextGenStripeRepository.php
jonwaldstein: Resolved 7f4ee705

jonwaldstein:
> @jonwaldstein I think we should capture these improvements somewhere, like a `@todo` or a new issue, so that they don't get lost in the merged/closed PR.

@kjohnson I created an issue in GiveWP #6686

jonwaldstein:
@kjohnson your feedback has been addressed.

kjohnson (approved):

PR #123, opened 2023-01-25
Feature: add compatibility for PayPal Standard

jonwaldstein:
Related: https://github.com/impress-org/givewp/pull/6684

## Description

This adds support for the PayPal standard gateway in our next gen forms.

This gateway has been implemented as a fully backwards compatible gateway while our next gen feature plugin is installed.  This means the PayPal standard gateway that is potentially already being used on legacy form - will still work there but will also now work on next gen forms.  There is nothing extra to enable or setup.

To achieve this, next gen is unregistering the core paypal gateway, and registering a new `PayPalStandardGateway` that extends the core one.  So all the same methods will work as expected on legacy, but next gen adds a few different conditionals for redirecting to the proper location and handling redirects.

## Affects

- The PayPal standard gateway, _only_ when next gen is installed.
- Introduces a new trait that can override what's in core `HandleHttpResponses` for our form's purposes. 
- Updates the form js to handle redirects properly

## Visuals

**Completing a payment**

**Cancelling a payment**

*Note: our form does not have a sense of flash messages yet, but at least we can bring them back to the form.

## Testing Instructions

- Setup PayPal standard in GiveWP (you'll need a developer & sandbox account)
- Donate using PayPal standard on a next gen form
- Donate using PayPal standard on a legacy form (make sure it works as expected)

**Note when using legacy form**
At the time of writing, the order of our gateway registration does not account for unregistering a gateway.
This results in seeing the paypal fields twice on the legacy form.  There is a [PR ](https://github.com/impress-org/givewp/pull/6684) in GiveWP core to resolve that issue.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:

src/NextGen/Framework/PaymentGateways/Traits/HandleHttpResponses.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function handleResponse($type)
+    {
+        if ($type instanceof RedirectResponse) {
+            if (isset($_SERVER['CONTENT_TYPE']) && str_contains($_SERVER['CONTENT_TYPE'], "application/json")) {
+                wp_send_json([
+                    'data' => [
+                        'redirectUrl' => $type->getTargetUrl()
+                    ]
+                ]);
```
JasonTheAdams, line 32: @jonwaldstein What do you think about adding response "types" for when we send JSON back to the browser? This way we can check the type and assume the shape of the corresponding data. Types might be:
- `redirect`
- `validation_error`
- `success`

Just a thought. 😄

src/NextGen/Gateways/PayPal/PayPalStandardGateway/PayPalLogo.tsx
```diff
@@ -0,0 +1,57 @@
+export default function PayPalLogo() {
```
JasonTheAdams, line 1: Please save this as an .svg file so its cachable and doesn't add bloat.

src/NextGen/Framework/PaymentGateways/Traits/HandleHttpResponses.php
jonwaldstein: I like that idea

src/NextGen/Gateways/PayPal/PayPalStandardGateway/payPalStandardGateway.tsx
```diff
@@ -0,0 +1,40 @@
...
+                rowGap: "1rem",
+                textAlign: "center"
+            }}>
+                <PayPalLogo/>
+                <b>{fields.heading}</b>
+                <div><b>{fields.subheading}:</b> {fields.body}</div>
+            </div>
+        );
+    }
+}
```
JasonTheAdams, line 38: This is a great example of a gateway that I think could live entirely in PHP. Whether it supports recurring could be inferred, and we can add a `supportedCurrencies()` method to the gateway which returns an array of supported currencies. Finally, the fields themselves aren't interactive, so the markup could be passed along directly and we use Interweave to display it safely.

JasonTheAdams:
Would still love your review on this, @kjohnson. More perspectives are even better as we form foundations.

src/NextGen/Gateways/PayPal/PayPalStandardGateway/payPalStandardGateway.tsx
jonwaldstein: Yeah I agree, Im thinking things like supportsRecurring and supportsCurrency can be localized from the gateway class so they don't have to add them here manually.  As for using interweave, i'm open to it but want to think through what a clear api for that would look like so gateways are not confused about the difference between JS fields and php rendered fields.

src/NextGen/Gateways/PayPal/PayPalStandardGateway/PayPalLogo.tsx
jonwaldstein: resolved b10b9bd3

src/NextGen/Framework/PaymentGateways/Traits/HandleHttpResponses.php
jonwaldstein: I added `redirect` and `success` types.  I'll circle back to validation when the time comes b3e69892

src/NextGen/DonationForm/resources/app/form/Form.tsx
```diff
@@ -57,16 +57,20 @@ const handleSubmitRequest = async (values, setError, gateway: Gateway) => {
...
             gatewayData: beforeCreatePaymentGatewayResponse,
         });
 
-        if (isRedirect) {
-            await handleRedirect(response);
+        if (isRedirectResponse) {
+            await handleRedirect(response.url);
+        }
+
+        if (response?.type === 'redirect') {
```
JasonTheAdams, line 72: @jonwaldstein This would be a good time to use a [Type Predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

jonwaldstein: @JasonTheAdams those are awesome! f95a46ed

I definitely plan on expanding these moving forward.  I think what I just added is a good start.

jonwaldstein:
Note: we are still deciding out how to handle billing details in our Next Gen forms.  So, currently the setting for _Billing Details_ in PayPal Standard settings is not functional.

PR #129, opened 2023-02-10
Refactor: disable update button until dirty

jonwaldstein:
## Description

Form Builder updates:
- Change "Publish" button to "Update"
- Add simple `isDirty` state and disable update button until `isDirty` is true.
- Clean up `BlockEditorContainer` and move InterfaceSkeleton into `BlockEditorInterfaceSkeletonContainer`
- Fix a few react warnings.
- Update currency component with types

Note:
- The `isDirty` state is a very basic implementation and is only set after updating the form.  Later we might decide to do an actual dynamic comparison of the previous / changed state to determine its dirty-ness.

## Affects

The form builder

## Testing Instructions

- pull branch and run `npm run build` (*note: I noticed running `npm run dev/watch:form-builder` instead of build would sometimes make the block provider not recognize the _first_ change in some cases which affects updating & dirty state.)
- change blocks/form settings and use the update button
- the update button should be disabled after using the update button.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:

packages/form-builder/src/types/formState.ts
```diff
@@ -7,4 +7,5 @@ import type {FormSettings} from './formSettings';
 export type FormState = {
     blocks: Block[];
     settings: FormSettings;
+    isDirty?: boolean;
```
kjohnson, line 10: We should consider introducing the concept of Editor State at some point in the next cycle. I think it would be better to delineate Form State as persisted state. By contrast, `isDirty` does not persist with the form.

kjohnson: There are also scattered lines of `useState` for things like sidebar visibility, etc, that could be included in Editor State.

kjohnson (approved):
Looks good to me.

As I mention in the comments, we should consider introducing the concept of Editor State (separate from Form State), but that isn't necessary to ship `v0.2.0`.

packages/form-builder/src/types/formState.ts
jonwaldstein: I agree

PR #140, opened 2023-03-09
Feature: Recurring Donations Promotion

kjohnson:
## Description

This PR adds the Recurring Donations Promotion when no active gateways support subscriptions (recurring).

## Visuals

image

## Testing Instructions

- Disable the (Next Gen) Test Gateway


packages/form-builder/src/common/getWindowData.ts
```diff
@@ -6,10 +6,21 @@ declare global {
...
-        };
+        },
+        formBuilderData?: {
+            gateways: Array<any>;
+        },
     }
 }
 
 export default function getWindowData() {
     return window.storageData;
```
kjohnson, line 17: The concept of storage data needs to be refactored as the scope has changed significantly over time.

kjohnson:
Based on @JasonTheAdams [feedback](https://www.loom.com/share/f8f4d7bc5f964e3a9a9425d1a420c148?highlightComment=11411304&t=195) on my weekly [screencast](https://www.loom.com/share/f8f4d7bc5f964e3a9a9425d1a420c148), we might want to consider updating this to take into account supported gateways that are inactive, which will prevent an misleading upsell and could add support for per-form gateways.

> I think the promo should only display if there are no gateways (active or inactive) that support recurring. Otherwise, we risk selling them on something they already have, but just don't have any recurring gateways active. I think that's a niche scenario, but valid nonetheless.
>
> It may be worth anticipating that Per Form Gateways will make this a thing within the Form Builder, as they check which gateways they want this specific form to allow.

packages/form-builder/src/common/getWindowData.ts
```diff
@@ -6,10 +6,21 @@ declare global {
             formDesigns: FormDesign[];
             formPage: FormPageSettings;
             currency: string;
-        };
+        },
+        formBuilderData?: {
+            gateways: Array<any>;
```
jonwaldstein, line 11: let's make a typescript `Gateway` type/interface with id, label, supportsSubscriptions.

Then here we can do `gateways: Gateway[]`

The other areas you are using this should then be aware of the properties of the gateway object.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
```diff
@@ -80,13 +82,25 @@ public function renderPage()
 
         $formBuilderStorage->loadInFooter()->enqueue();
 
+        $builderPaymentGatewayData = array_map(function ($gateway) {
+            return [
+                'id' => $gateway::id(),
+                'label' => give_get_gateway_checkout_label($gateway::id()) ?? $gateway->getPaymentMethodLabel(),
+                'supportsSubscriptions' => $gateway->supportsSubscriptions(),
+            ];
+        }, give(DonationFormRepository::class)->getEnabledPaymentGateways($donationFormId));
```
jonwaldstein, line 91: Remember we also need to filter these gateways by next gen compatibility or _all_ gateways will be loaded. We should be able to just check if they implement `NextGenPaymentGatewayInterface`

packages/form-builder/src/promos/recurring-donations.tsx
```diff
@@ -0,0 +1,29 @@
...
+            boxShadow: '0 2px 2px 0 rgba(221, 221, 221, 0.25)',
+            border: 'solid 0.5px var(--givewp-primary-700)'
+        }}
+    >
+        <div style={{}}>
+            {__('Provide donors the option of making flexible recurring donations.', 'give')}
+        </div>
+        <div style={{display: 'flex', justifyContent: 'space-around'}}>
+            <Button href='#' variant={'primary'} style={{backgroundColor:'var(--givewp-primary-500)', padding: '4px 8px', height: 'auto'}}>Upgrade your plan</Button>
+            <Button href='#' variant={'link'} style={{color: 'var(--givewp-gray-100)'}}>Read more</Button>
```
jonwaldstein, line 23: Let's update these links to go to `https://givewp.com/addons/recurring-donations` for now. (open in new tab)

jonwaldstein (changes requested):

packages/form-builder/src/promos/recurring-donations.tsx
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/140/commits/9279878afb6150856c255a3475d2c965c9a55ffc.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/140/commits/3ab5ba6f2e9690ee68b62160d72d3ad65582ffcf.

packages/form-builder/src/common/getWindowData.ts
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/140/commits/936806a7864c0237645d3d4f813d8a76413cda93.

kjohnson:
@jonwaldstein this is ready for review.

jonwaldstein (approved):
Great job!

PR #145, opened 2023-03-28
Fix: Add fallback value for setPrice when fixed

kjohnson:
## Description

This PR ensures a fallback value for the `setPrice` value when using a fixed donation.

## Visuals

Peek 2023-03-28 09-19

## Testing Instructions

- Set the Donation Option to a Fixed Donation.
- Delete the value of the Set Donation setting.
- The Set Donation should fallback to `100`.

---
  - https://app.asana.com/0/0/1203940464851369

### DIFF
--- a/packages/form-builder/src/blocks/fields/amount/inspector/index.tsx
+++ b/packages/form-builder/src/blocks/fields/amount/inspector/index.tsx
@@ -29,7 +29,7 @@ const Inspector = ({attributes, setAttributes}) => {
                     <CurrencyControl
                         label={__('Set Donation', 'give')}
                         value={setPrice}
-                        onBlur={() => setPrice || setAttributes({setPrice: 100})}
+                        onBlur={() => setPrice || setAttributes({setPrice: 25})}
                         onValueChange={(setPrice) => setAttributes({setPrice: setPrice ? parseInt(setPrice) : 0})}
                     />
                 )}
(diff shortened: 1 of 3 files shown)


packages/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -29,7 +29,8 @@ const Inspector = ({attributes, setAttributes}) => {
                     <CurrencyControl
                         label={__('Set Donation', 'give')}
                         value={setPrice}
-                        onValueChange={(setPrice) => setAttributes({setPrice})}
+                        onBlur={() => setPrice || setAttributes({setPrice: 100})}
```
jonwaldstein, line 32: I just checked legacy and our default value there is 25.  Can we update to that instead so at least it matches current functionality?

Screenshot 2023-03-28 at 4 21 45 PM

jonwaldstein (changes requested):

packages/form-builder/src/blocks/fields/amount/inspector/index.tsx
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/145/commits/c776c8797807b40ab7d33d2a615e7f5beddc7970.

jonwaldstein (approved):
👍

PR #147, opened 2023-03-28
Refactor: Set form status as draft by default

kjohnson:
## Description

This PR sets the default status for a new form as "draft", with the form being published on save.

## Visuals

| Status: Publish |
| --- |
| image |

| Status: Draft |
| --- |
| image |

| All Forms Table |
| --- |
| image |

## Testing Instructions

- Create a new form and check that the default status is DRAFT.
- Publish the form and check that the status is PUBLISH.
- Switch the form to a draft.

### DIFF
--- a/src/FormBuilder/ViewModels/FormBuilderViewModel.php
+++ b/src/FormBuilder/ViewModels/FormBuilderViewModel.php
@@ -36,7 +36,7 @@ class FormBuilderViewModel
             }, give(FormDesignRegistrar::class)->getDesigns()),
             'formPage' => [
                 'isEnabled' => give_is_setting_enabled(give_get_option('forms_singular')), // Note: Boolean values must be nested in an array to maintain boolean type, see \WP_Scripts::localize().
-                'permalink' => add_query_arg(['p' => $donationFormId], site_url()),
+                'permalink' => add_query_arg(['p' => $donationFormId], site_url('?post_type=give_forms')),
                 'rewriteSlug' => get_post_type_object('give_forms')->rewrite['slug'],
             ],
         ];


packages/form-builder/src/containers/HeaderContainer.tsx
```diff
@@ -85,12 +88,25 @@ const HeaderContainer = ({
...
                         disabled={isSaving || !isDirty}
                         variant="primary"
                     >
-                        {isSaving ? __('Updating...', 'give') : __('Update', 'give')}
+                        {isSaving
+                            ? __('Updating...', 'give')
+                            : 'publish' === formSettings.formStatus
+                                ? __('Update', 'give')
+                                : __('Publish', 'give')
+                        }
```
jonwaldstein, line 109: I just want to point out a few things here that I noticed.

1. When you first create the form, it's in draft but the publish button is disabled.  We need to make sure publish can _always_ be clicked when in draft mode (regardless if dirty) so people can always publish the form when satisfied.  Then we can go back to checking dirty state if published on the update button. 
2. Clicking "save as draft" should modify the test to `saving...` but not the modify the publish button text like below:

Screenshot 2023-03-31 at 10 25 12 AM

3. When a form is in draft mode - we should be able to view the form on the front-end when logged in (That's how native WP posts/post-types work by using the postID instead of the slug)

jonwaldstein (changes requested):

packages/form-builder/src/containers/HeaderContainer.tsx
kjohnson: I updated the draft and publish button text and disabled states to account for the different circumstances, see https://github.com/impress-org/givewp-next-gen/pull/147/commits/bf00d2a34574a719a151ad844ea2f92383fbedf8.

kjohnson: I confirmed that draft forms can be viewed at the permalink when authenticated - and that draft forms have a 404 redirect when unauthenticated.

jonwaldstein (approved):
👍

PR #155, opened 2023-04-04
Chore: update implementation of react render to support react 18 & WordPress 6.2

jonwaldstein:
## Description

The donation form uses `@wordpress/element` and related packages for react to play nicely with other blocks. It appears newer versions of WordPress are using react 18 where the root rendering api is a little different.  WP [docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-element/) have a conditional to provide backwards compatibility so this PR updates our implementation to follow that approach.

>Note: createRoot was introduced with React 18, which is bundled with WordPress 6.2. Therefore it may be necessary to mount your component depending on which version of WordPress (and therefore React) you are currently using. This is possible by checking for an undefined import and falling back to the React 17 method of mounting an app using render.

## Affects

The donation form and confirmation page rendering.

## Visuals

N/A

## Testing Instructions

- Run `npm run dev` and make sure the form and confirmation load properly.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/NextGen/DonationForm/resources/app/DonationFormApp.tsx
+++ b/src/NextGen/DonationForm/resources/app/DonationFormApp.tsx
@@ -1,4 +1,4 @@
-import {render} from '@wordpress/element';
+import {createRoot, render} from '@wordpress/element';
 import getDefaultValuesFromSections from './utilities/getDefaultValuesFromSections';
 import Form from './form/Form';
 import {GiveDonationFormStoreProvider} from './store';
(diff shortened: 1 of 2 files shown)


kjohnson:
I found this explanation for the change, which is interesting.

> ## What are the differences?
>
> There are a few reasons we changed the API.
>
> First, this fixes some of the ergonomics of the API when running updates. As shown above, in the legacy API, you need to continue to pass the container into render, even though it never changes. This also mean that we don’t need to store the root on the DOM node, though we still do that today.
>
> Second, this change allows us to remove the hydrate method and replace with with an option on the root; and remove the render callback, which does not make sense in a world with partial hydration.
>
> https://github.com/reactwg/react-18/discussions/5

kjohnson (approved):

PR #159, opened 2023-04-06
Feature: add backwards compatibility for legacy Stripe webhooks

jonwaldstein:
Depends on (but not completely): https://github.com/impress-org/givewp/pull/6768

## Description

We discovered that (currently), one of the existing Stripe gateways must be enabled in order for the Stripe webhooks to run.  This adds our new Stripe gateway to that list using a new filter `give_stripe_supported_payment_methods`.

Also included is a direct call to the legacy class to include the necessary files if this filter is not being used.  It checks if the `Give_Stripe_Webhooks` exists and if the new Stripe gateway is enabled.  This will make this backwards compatible regardless of the filter.

## Affects

- This will make our new `NextGenStripeGateway` not have to rely on a legacy Stripe gateway to be enabled for webhooks to work properly.

## Visuals

N/A

## Testing Instructions

- Test the Stripe webhook with only the next gen gateway enabled.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/give-next-gen.php
+++ b/give-next-gen.php
@@ -9,6 +9,7 @@ use Give\NextGen\CustomFields\ServiceProvider as CustomFieldsServiceProvider;
 use Give\NextGen\DonationForm\ServiceProvider as DonationFormServiceProvider;
 use Give\NextGen\FormPage\ServiceProvider as FormPageServiceProvider;
 use Give\NextGen\Framework\FormDesigns\ServiceProvider as FormDesignServiceProvider;
+use Give\NextGen\Gateways\Stripe\LegacyStripeAdapter;
 use Give\NextGen\ServiceProvider as NextGenServiceProvider;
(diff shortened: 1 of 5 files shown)


jonwaldstein:

jonwaldstein:
I [commented](https://github.com/impress-org/givewp/pull/6768#issuecomment-1499702619) on the PR this depends on. 
 It appears this is not working at the moment when the function is called from GiveWP.

jonwaldstein:
@kjohnson this is ready for review again.  I have it setup so the legacy Stripe Webhooks will get loaded regardless of the new filter.  I still like the idea of using the filter so we can completely deviate from legacy Stripe by simply removing it from the list.  This will give us both options & backwards compatibility with GiveWP <= 2.26.0.

kjohnson (approved):
Works for me, locally. Let's get this up on a staging environment for some final testing.

PR #164, opened 2023-04-11
Fix: Correct reference to Stripe Next Gen gateway when loading legacy Stripe code

kjohnson:
## Description

As reported by @Benunc, the Next Gen feature plugin was causing a break in form styles for the (legacy) Multi-Part template. I narrowed down the issue to the Next Gen Test Gateway and the (legacy) Stripe front-end files, which are included for the webhook listeners.

The `loadLegacyStripeWebhooksAndFilters()` method was checking for the `NextGenTestGateway::class` to be "active", but should instead have been checking for the `NextGenStripeGateway::class` - a simple typo in the code and oversight during code review.

Additionally, there issue revealed a bug in the legacy Stripe code, in which a callback for the filter hook `give_form_html_tags` returned `false` if no Stripe gateways were active when the filter hook expects an array to always be returned.

## Visuals

| Broken | Fixed |
| --- | --- |
image | image |

## Testing Instructions

Using a (legacy) Multi-Part form template:

- Toggle the Next Gen _Test_ Gateway
- Check the form field styles

---
  - https://app.asana.com/0/0/1204358626491010

### DIFF
--- a/src/NextGen/Gateways/Stripe/LegacyStripeAdapter.php
+++ b/src/NextGen/Gateways/Stripe/LegacyStripeAdapter.php
@@ -15,6 +15,7 @@ class LegacyStripeAdapter
      * This makes it possible to load the files without having to enable a legacy stripe gateway.
      * This also makes it possible to load the files without the use of the give_stripe_supported_payment_methods filter.
      *
+     * @unreleased Fix reference to Next Gen gateway when loading Stripe (legacy) code.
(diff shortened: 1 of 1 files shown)


JasonTheAdams (approved):
Good find, @kjohnson! Please shoot this over to QA for a quick test.

Benunc (approved):
appears to work now! :shipit:

PR #167, opened 2023-04-14
Fix: Move editor out from under header in the Custom Styles modal

kjohnson:
## Description

This PR removes the top margin offset from the code editor in the Custom Styles modal so that the editor is no longer hidden by the header.

## Visuals

| After |
| --- |
|image|

| Before |
| --- |
| image |

## Testing Instructions

 
In the Design tab, under the Custom Styles section, open the custom styles modal.


JasonTheAdams (approved):
Much better, thank you!

PR #179, opened 2023-05-08
Feature: add support for form builder extensions

pauloiankoski:
## Description

This pull request adds initial support for Form Extensions, such as actions requiring scripts and styles to be enqueued, new categories for blocks, and a registrar to register custom blocks for the Form Builder.

**Enqueue Form Builder Scripts**

```
do_action('givewp_form_builder_enqueue_scripts')
```

**Register Block**

```
window.givewp.form.blocks.register(blockName, blockSettings);
```

**Enqueue Donation Form Scripts**

```
do_action('givewp_donation_form_enqueue_scripts');
````

Also, a new provider was added for interacting with the DonationSummary.

**Interact with DonationSummary using new hook** 

```
const {useDonationSummary} = window.givewp.form.hooks;
const donationSummary = useDonationSummary();

donationSummary.addItem(item);
donationSummary.removeItem(itemId);
donationSummary.addToTotal(itemId, amount);
donationSummary.removeFromTotal(itemId)
```

## Pre-review Checklist

-   [x] Includes unit tests

### CODE REVIEW

File: src/NextGen/DonationForm/ViewModels/DonationFormViewModel.php [hunk 1]
```diff
@@ -269,6 +271,29 @@ private function enqueueFormScripts(int $formId, string $formDesignId)
             }
         }
 
+        // load extensions
```
kjohnson, line 274: Is this duplicated from `src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php`?

src/NextGen/Framework/FormExtensions/Registrars/FormExtensionRegistrar.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function registerExtension(string $extensionClass)
+    {
+        try {
+            $this->register($extensionClass);
+        } catch (InvalidArgumentException $invalidArgumentException) {
+            Log::error('Form Extension Registration', ['data' => $invalidArgumentException->getMessage()]);
+            throw $invalidArgumentException;
+        } catch (OverflowException $overflowException) {
+            Log::error('Form Extension Registration ', ['data' => $overflowException->getMessage()]);
```
kjohnson, line 64: These log messages should probably be more specific, explaining what went wrong.

src/NextGen/Framework/FormExtensions/Registrars/FormExtensionRegistrar.php
```diff
@@ -0,0 +1,108 @@
...
+            throw $overflowException;
+        }
+    }
+
+    /**
+     * @unreleased 
+     */
+    public function unregisterExtension(string $extensionId)
+    {
+        if (isset($this->extensions[$extensionId])) {
```
kjohnson, line 74: I would suggest leveraging `hasExtension()` instead of using the `isset` directly.

src/NextGen/Framework/FormExtensions/ServiceProvider.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @inheritDoc
+     */
+    public function register()
+    {
+        give()->singleton(FormExtensionRegistrar::class, function () {
+            $registrar = new FormExtensionRegistrar();
+
+            Hooks::doAction('givewp_register_form_extension', $registrar);
```
kjohnson, line 22: Running `do_action` inside of the `register()` method may be too soon in the setup process. Instead we should probably be triggering hooks in the `boot()` method.

packages/form-builder/src/index.tsx
```diff
@@ -5,10 +5,13 @@ import {BlockSupports, registerBlockType} from '@wordpress/blocks';
...
 
 import sectionBlocks, {sectionBlockNames} from './blocks/section';
-import fieldBlocks from './blocks/fields';
-import elementBlocks from './blocks/elements';
+import blockRegistrar from '@givewp/form-builder/common/registrars/blocks';
 import {FieldBlock} from '@givewp/form-builder/types';
 
+import './blocks/fields';
+import './blocks/elements';
+import './blocks/extensions';
```
kjohnson, line 13: Imports should return something. Otherwise this results in side effects, which can have unintended consequences.

pauloiankoski: @kjohnson I've refactored the blocks registrar in a way that handles all the registration processes, including the `registerBlockType` step. It looks much clean now.

See: https://github.com/impress-org/givewp-next-gen/pull/179/files/59e2f3d8cfe0df6f944bbe272fd069d30189dbae..adf55856f6fec2a27d7fd123705e116e3fd7a7b5

src/NextGen/Framework/FormExtensions/ServiceProvider.php
pauloiankoski: Changed it on 4a407d9d52e2d4c1591de3afc332b129c0b62272 then tested it. Still works fine.

src/NextGen/DonationForm/ViewModels/DonationFormViewModel.php
pauloiankoski: It's not exactly a duplicate as we call distinct methods based on the current screen.

`RegisterFormBuilderPageRoute.php` uses:
```php
$extension->formBuilderCss();
$extension->formBuilderJs();
$extension->formBuilderDependencies();
```

`DonationFormViewModel.php` uses:
```php
$extension->formViewCss();
$extension->formViewJs();
$extension->formViewDependencies();
```

src/NextGen/Framework/FormExtensions/Registrars/FormExtensionRegistrar.php
pauloiankoski: Good call. Changed on 4f64f56a35a6bda20b7898d4dd30e7651483d65a

packages/form-builder/src/index.tsx
```diff
@@ -1,45 +1,24 @@
...
-        supports: {
-            ...settings.supports,
-            ...supportOverrides,
-        },
-    })
-);
+blockRegistrar.register(sectionBlocks);
+blockRegistrar.register(fieldBlocks, sectionBlockNames);
+blockRegistrar.register(elementBlocks, sectionBlockNames);
+blockRegistrar.register(extensionBlocks, sectionBlockNames);
```
jonwaldstein, line 15: @pauloiankoski this is looking great.  I think we're going to want to elevate the `blockRegistrar` to the window and have it load as a standalone script _before_ the form builder and add-ons.  

This would mean both add-ons and the form-builder would do this instead:

`window.givewp.blocks.registrar.register()`

I would even consider registering our core blocks before add-ons to prevent any conflicts.  That could be a nice seperation between our blocks and the form builder instance itself - which is how it's done in wp core.

src/NextGen/Framework/FormExtensions/ServiceProvider.php
```diff
@@ -0,0 +1,34 @@
...
+class ServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function register()
+    {
+        give()->singleton(FormExtensionRegistrar::class, function () {
+            return new FormExtensionRegistrar();
+        });
```
JasonTheAdams, line 21: You don't need to use a callback unless you're doing something special when its instantiated:
```php
give()->singleton(FormExtensionRegistrar::class);
```

PR #181, opened 2023-05-10
Feature: Per Form Email Settings

kjohnson:
## Description

This PR adds per form email settings to the Visual Donation Form Builder, which includes per form email template settings

The email template settings are dynamically generated to support any registered email notification types.

The email template preview is duplicated and refactored to support passing dynamic preview values.

- Adds `templateTags` to `window.storageData` with corresponding `TemplateTag` type.
- Adds `emailNotifications` to `window.storageData` with corresponding `EmailNotification` type.
- Adds the `useClipboard` hook from `react-use-clipboard`.
- Adds  'templateTags', 'emailNotifications' , and 'emailPreviewURL' as localized data on the Page Route.
- Adds the REST API namespace `givewp/next-gen/email-preview` for the `/show`, `/send` routes.

## Affects

`src/FormBuilder/Actions/DequeueAdminScriptsInFormBuilder.php`
- The `DequeueAdminScriptsInFormBuilder` action is basically disabled, so that the dependencies for the Media Library are included, but this could be more nuanced and not so heavy handed.

## Backwards Compatibility

Per form email settings and per form email template options are propagated to their respective backwards compatible form meta so that the legacy email system continues to work with these changes.

- `\Give\FormBuilder\Actions\UpdateEmailSettingsMeta`
- `\Give\FormBuilder\Actions\UpdateEmailTemplateMeta`

## Visuals

image

image

Peek 2023-05-30 12-12

image

## Testing Instructions

Create a form, modify per form email settings, view and send preview emails.


src/FormBuilder/Actions/DequeueAdminScriptsInFormBuilder.php
```diff
@@ -22,7 +22,7 @@ public function __invoke()
         if ($this->isFormBuilderRoute()) {
             $wpScripts = wp_scripts();
 
-            $wpScriptsRegistered = array_column($wpScripts->registered, 'handle');
+            $wpScriptsRegistered = []; //array_column($wpScripts->registered, 'handle');
```
kjohnson, line 25: We need some core scripts for the Media Library integration to work. This needs to be revisited so we don't just enqueue everything.

packages/form-builder/src/settings/email/template-options/settings.tsx
```diff
@@ -0,0 +1,182 @@
...
+                        label={__('Email content type', 'givewp')}
+                        help={__('Choose email type', 'givewp')}
+                        value={option.content_type || 'html'}
+                        options={[
+                            {label: __('HTML', 'givewp'), value: 'html'},
+                            {label: __('Plain', 'givewp'), value: 'plain'},
+                        ]}
+                    />
+
+                    <TextControl
```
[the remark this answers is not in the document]
kjohnson: ## New Donation (Admin)

image

## Donation Receipt (Donor)

image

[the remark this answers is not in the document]
kjohnson: This is new discovery and I think should be a separate PR.

packages/form-builder/src/common/getWindowData.ts
```diff
@@ -13,6 +22,8 @@ declare global {
                 isInstalled: boolean;
             },
             gatewaySettingsUrl: string;
+            templateTags: TemplateTag[];
```
jonwaldstein, line 25: I think we should rename this to be specific to emails like `emailTemplateTags` - reason being, there are other settings like the thank you page that have their own.

packages/form-builder/src/types/formSettings.ts
```diff
@@ -22,4 +23,10 @@ export type FormSettings = {
     receiptHeading: string;
     receiptDescription: string;
     formStatus: FormStatus;
+    emailTemplateOptions: EmailTemplateOption[];
+    emailOptionsStatus: string;
+    emailTemplate: string;
+    emailLogo: string;
+    emailFromName: string;
+    emailFromEmail: string;
```
jonwaldstein, line 31: As we get into these settings with a lot of options, we might consider nesting them into their own object to simplify things like:

```
formSettings: {
  email: {
    templateOptions,
    optionStatus,
    template,
    logo,
    fromName,
    fromEmail
  }
}
```

src/FormBuilder/Controllers/FormBuilderResourceController.php
```diff
@@ -77,6 +79,9 @@ public function update(WP_REST_Request $request)
             return rest_ensure_response($requiredFieldsError);
         }
 
+        give(UpdateEmailSettingsMeta::class)->__invoke($form);
+        give(UpdateEmailTemplateMeta::class)->__invoke($form);
+
```
jonwaldstein, line 84: any reason we can't use the invokable syntax?

`(new UpdateEmailSettingsMeta())($form)`

src/FormBuilder/DataTransferObjects/EmailNotificationData.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /** @var bool */
+    public $supportsRecipients;
+
+    /** @var array */
+    public $defaultValues;
+
+    public static function fromLegacyNotification($notification): EmailNotificationData
+    {
+        return (new ConvertNotificationToDTO($notification))->toDto();
```
jonwaldstein, line 29: If this is specific to legacy we should rename to `ConvertLegacyNotificationToEmailNotificationData`.

Also in light of being an action this should not need to call `toDto()` and instead by invoking should return the DTO.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
```diff
@@ -111,6 +112,11 @@ public function renderPage()
                 'isInstalled' => defined('GIVE_RECURRING_VERSION'),
             ],
             'gatewaySettingsUrl' => admin_url('edit.php?post_type=give_forms&page=give-settings&tab=gateways'),
+            'templateTags' => array_values(give()->email_tags->get_tags()),
+            'emailNotifications' => array_map(static function ($notification) {
+                return EmailNotificationData::fromLegacyNotification($notification);
+            }, apply_filters('give_email_notification_options_metabox_fields', array(), $donationFormId)),
+            'emailPreviewURL' => rest_url('givewp/next-gen/email-preview'),
```
jonwaldstein, line 119: This object is starting to get big enough where we can move into the viewModel so we can write a unit test for it's shape.

`$formBuilderViewModel->getformBuilderData()`

src/FormBuilder/Actions/UpdateEmailSettingsMeta.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Give\FormBuilder\Actions;
+
+use Give\NextGen\DonationForm\Models\DonationForm;
+
+/**
+ * Update email settings on backwards compatible form meta.
+ */
+class UpdateEmailSettingsMeta
+{
+    public function __invoke(DonationForm $form)
```
jonwaldstein, line 12: please add `@unreleased`

src/FormBuilder/Actions/UpdateEmailTemplateMeta.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Give\FormBuilder\Actions;
+
+/**
+ * Update email template options on backwards compatible form meta.
+ */
+class UpdateEmailTemplateMeta
+{
+    public function __invoke($form)
```
jonwaldstein, line 10: please add `@unreleased`

src/FormBuilder/DataTransferObjects/EmailNotificationData.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Give\FormBuilder\DataTransferObjects;
+
+use Give\FormBuilder\Actions\ConvertNotificationToDTO;
+
+/**
+ * @since 0.1.0
```
jonwaldstein, line 8:
```suggestion
 * @unreleased
```

src/NextGen/EmailPreview/Actions/ApplyPreviewTemplateTags.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace Give\NextGen\EmailPreview\Actions;
+
+class ApplyPreviewTemplateTags
+{
+    public function __invoke($message)
```
jonwaldstein, line 7: please add `@unreleased`

src/NextGen/EmailPreview/Actions/BuildEmailPreview.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Give\NextGen\EmailPreview\Actions;
+
+class BuildEmailPreview
+{
+    public function __construct(ApplyPreviewTemplateTags $applyPreviewTemplateTagsAction)
```
jonwaldstein, line 7: please add `@unreleased`

src/NextGen/EmailPreview/Actions/GetEmailNotificationByType.php
```diff
@@ -0,0 +1,24 @@
...
+
+class GetEmailNotificationByType
+{
+    /**
+     * @param $type
+     *
+     * @return \Give_Email_Notification|void
+     * @throws \Exception
+     */
+    public function __invoke($type)
```
jonwaldstein, line 13: please add `@unreleased`

src/NextGen/EmailPreview/Controllers/SendEmailPreview.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Give\NextGen\EmailPreview\Controllers;
+
+use Give\NextGen\EmailPreview\Actions\BuildEmailPreview;
+
+class SendEmailPreview
+{
+    public function __invoke(\WP_REST_Request $request)
```
jonwaldstein, line 9: please add `@unreleased`

src/NextGen/EmailPreview/Controllers/SendEmailPreview.php
```diff
@@ -0,0 +1,20 @@
...
+    {
+        $email = $request->get_param('email_address');
+        $subject = $request->get_param('email_subject');
+        $message = give(BuildEmailPreview::class)->__invoke($request);
+
```
jonwaldstein, line 14:
```suggestion
        $message = (new BuildEmailPreview())($request);

```

src/NextGen/EmailPreview/Controllers/ShowEmailPreview.php
```diff
@@ -0,0 +1,18 @@
...
+
+class ShowEmailPreview
+{
+    public function __invoke(\WP_REST_Request $request)
+    {
+        Give()->emails->__set('html', true); // Show formatted text in browser even text/plain content type set for an email.
+
+        ob_clean();
+        header('Content-Type: text/html; charset=UTF-8');
+        echo give(BuildEmailPreview::class)->__invoke($request);
```
jonwaldstein, line 15:
```suggestion
        echo (new BuildEmailPreview())($request);
```

src/NextGen/EmailPreview/ServiceProvider.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /*
+     * @inheritdoc
+     */
+    public function boot()
+    {
+        add_action('rest_api_init', [$this, 'registerRoutes']);
+    }
+
+    public function registerRoutes()
```
jonwaldstein, line 25: please add `@unreleased`

packages/form-builder/src/settings/email/template-options/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    const [ showPreview, setShowPreview ] = useState<boolean>( false );
+
+    const {templateTags, emailNotifications} = getFormBuilderData();
+
+    const EmailPreviewContent = ({emailType}) => {
+
+        const [ previewHtml, setPreviewHtml ] = useState<string>(null);
+
+        const {settings: {emailTemplateOptions, emailTemplate, emailLogo, emailFromName, emailFromEmail}} = useFormState();
```
jonwaldstein, line 28: you'll want to update the type `FormState` to account for new settings

packages/form-builder/src/settings/email/template-options/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const openModal = () => setOpen( true );
+    const closeModal = () => setOpen( false );
+
+    const [ selectedTab, setSelectedTab ] = useState<string>();
+
+    const [ showPreview, setShowPreview ] = useState<boolean>( false );
+
+    const {templateTags, emailNotifications} = getFormBuilderData();
+
+    const EmailPreviewContent = ({emailType}) => {
```
jonwaldstein, line 24: Does this need to be a nested component? seems like it should be pulled out into it's own, especially because it has some side effects.

src/FormBuilder/Controllers/FormBuilderResourceController.php
```diff
@@ -77,6 +79,9 @@ public function update(WP_REST_Request $request)
...
         $form->status = $updatedSettings->formStatus;
         $form->save();
```
jonwaldstein, line 87: Another note, I think we should just add an action here, `do_action('givewp_form_builder_updated', $form)` that we can add these legacy actions to so they're not hard-coded here.

jonwaldstein (changes requested):
@kjohnson this seems to be working well.  Added some suggestions.

Also, I was not getting the right email formatting during preview, tried plain text and html and still getting this html blob:

<details>
Screenshot 2023-05-15 at 4 58 32 PM
</details>

kjohnson:
@jonwaldstein Re: Preview Formatting: The HTML formatting is forced during preview, which was a decision carried over from the legacy code that I was adapting.

See  [Fix: reset email html and template params](https://github.com/impress-org/givewp/commit/0a38a636335b406aff4254754862f3cbde81fa95):
```diff
Give()->emails->__set( 'content_type', $content_type );
+ Give()->emails->__set( 'html', true );
```

That said, I'm open to changing this decision and now would be the time for such a change.

packages/form-builder/src/common/getWindowData.ts
kjohnson: Good point - I'll refactor the name.

src/FormBuilder/Controllers/FormBuilderResourceController.php
kjohnson: Either method works, but I prefer resolving from the container, which is mock-able and supports dependency injection in the invoke-able class (neither of which is currently being used).

Otherwise, I would have reached for the invokable syntax if I needed to define the constructor arguments manually, which is not the case here.

src/FormBuilder/DataTransferObjects/EmailNotificationData.php
kjohnson: Works for me - will do.

packages/form-builder/src/settings/email/template-options/index.tsx
kjohnson: There may have been a specific reason, or it could just be an artifact of a refactor. I'll confirm either way and extract if it doesn't need to be nested.

src/FormBuilder/Controllers/FormBuilderResourceController.php
kjohnson: What would be the difference between adding these here versus elsewhere? There isn't another location in the code base that comes to mind without creating a new section just to replicate the above.

Regardless, I can see that the `do_action` would beneficial for extensibility in add-ons.

src/FormBuilder/Controllers/FormBuilderResourceController.php
jonwaldstein: Gotcha, in general I prefer to call actions outside of the container since they're meant to be isolated in nature.

packages/form-builder/src/settings/email/template-options/index.tsx
kjohnson: These have been added to the `FormSettings` type, which is used in `FormState`.

kjohnson:
@jonwaldstein I've pushed updates for the requested changes, except for a couple that I would like to make as a separate PR.

packages/form-builder/src/types/formSettings.ts
kjohnson: I started working through this, but in practice it felt more complicated and hard to read when using the settings.

kjohnson:
@jonwaldstein this is ready for re-review.

src/FormBuilder/Actions/DequeueAdminScriptsInFormBuilder.php
```diff
@@ -22,6 +22,8 @@ public function __invoke()
         if ($this->isFormBuilderRoute()) {
             $wpScripts = wp_scripts();
 
+            $wpScriptsRegistered = array_column($wpScripts->registered, 'handle');
+
```
jonwaldstein, line 26: (from merge) This can be removed

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
```diff
@@ -77,41 +80,32 @@ public function renderPage()
...
+            GIVE_NEXT_GEN_URL,
+            'give'
+        ))
+            ->dependencies(
+                $this->getRegisteredFormBuilderJsDependencies(
+                    $formBuilderViewModel->jsDependencies()
+                )
+            )
+            ->loadInFooter()
+            ->enqueue();
```
jonwaldstein, line 108: (from merge) switching to `wp_enqueue_script` was intentional here and should be reverted

src/FormBuilder/ServiceProvider.php
```diff
@@ -36,7 +39,7 @@ public function boot()
 
         Hooks::addAction('admin_menu', RegisterFormBuilderPageRoute::class);
 
-        Hooks::addAction('admin_print_scripts', DequeueAdminScriptsInFormBuilder::class);
+//        Hooks::addAction('admin_print_scripts', DequeueAdminScriptsInFormBuilder::class);
```
jonwaldstein, line 43: This can be uncommented as it's only dequeuing specific givewp legacy scripts

tests/Unit/ViewModels/FormBuilderViewModelTest.php
```diff
@@ -29,31 +29,37 @@ public function testShouldReturnStorageData()
...
+            'formPage',
+            'gateways',
+            'gatewaySettingsUrl',
+            'isRecurringEnabled',
+            'recurringAddonData',
+            'emailTemplateTags',
+            'emailNotifications',
+            'emailPreviewURL',
+        ] as $key) {
+            $this->assertArrayHasKey($key, $storageData);
```
jonwaldstein, line 52: This is a pretty big difference in testing as before we were making sure the entire array keys and values are expected - where as now we're only checking the array key.  Can you change this back to `assertSame`?

jonwaldstein (changes requested):

packages/form-builder/src/App.scss
```diff
@@ -2,6 +2,603 @@
...
+    // Layout
+    --givewp-sidebar-width: 284px;
+
+    // Primary
+    --givewp-primary-200: #e6ffe6;
+    --givewp-primary-300: #c2f2c3;
+    --givewp-primary-400: #aaf2ac;
+    --givewp-primary-500: #68bf6b;
+    --givewp-primary-600: #15ae56;
+    --givewp-primary-700: #5ba65e;
```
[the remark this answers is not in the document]
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/181/commits/34c4e5af502a95a648c8e29d41ee1417a6ebc8e9.

src/FormBuilder/Actions/DequeueAdminScriptsInFormBuilder.php
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/181/commits/01b226ccb13828fd75e51437234609c4cf70c562.

tests/Unit/ViewModels/FormBuilderViewModelTest.php
kjohnson: Asserting the values in addition to the shape feels heavy handed and tightly couples the test to the implementation. Any change to the values or any addition of filters would break a fragile test like this.

My understanding is that we should be testing the _shape_ of the data, which asserting the keys accomplishes.

jonwaldstein: When it comes to writing tests for view models, I prefer to be heavy handed so I can be confident the data given to the view is exactly what it needs to function.  In fact, that's one of the main reasons I create view models - to extract predictable data I can write tests for.  Our front-end is fragile in nature, relying on a very specific data shape with specific data values and if either one of those is off, the view won't work properly and could break the page. 

So I see changes to the values affecting the tests a huge benefit and in my experience has prevented the front-end from breaking.  If somewhere up the chain there is a discrepancy between one of those values - and the front-end is not expecting it - i'm hopeful the test would catch it first.

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
kjohnson: Ok - I'm following now. I'll get this documented for future us.

kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/181/commits/f902db52916b8ca161c761e9450ac83c1aa29b09, https://github.com/impress-org/givewp-next-gen/pull/181/commits/7f34b404ba18ca1a51fba5f18b5c4655ac0fefb9

tests/Unit/ViewModels/FormBuilderViewModelTest.php
kjohnson: Ok, so you are wanting the redundancy here as supporting a critical system. I'll revert my changes to the test.

kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/181/commits/ed71b41a7652ee079df6cfbd7ff58d366b6aa7f1

src/FormBuilder/Routes/RegisterFormBuilderPageRoute.php
jonwaldstein: @kjohnson thanks man!

tests/Unit/ViewModels/FormBuilderViewModelTest.php
jonwaldstein: Yup, well said. Much appreciated

jonwaldstein:
@kjohnson one weird little quirk I noticed is when you create a new form, click into customize email settings and immediately try sending a test email with "global option" set - i'm not receiving the email, same with previewing.  Until you select "customize" within a tab it starts to work.  Is this happening to you?

Screenshot 2023-05-23 at 7 22 55 PM

kjohnson:
I updated the Email Preview Panel to be conditionally shown if the selected email is set to be customized.

Peek 2023-05-30 12-12

jonwaldstein:
@kjohnson I think that change is affecting some of the email settings.  Two issues I just noticed
- Preview email is not respecting the default/selected content type until its changed
- Send test email is not sending the subject or email header - just the content

Can you play around with the functionality of the settings please.

kjohnson:
Ah, I see the issue with the default values. Since the server was originally forcing `text/html` for the preview I didn't notice that the default was not being set correctly.

jonwaldstein:
@kjohnson couple more small things I noticed but will approve in the meantime as they are very minor:

**The quill link inserter gets chopped off:**

Screenshot 2023-06-01 at 3 17 55 PM

**The tag descriptions are not converting html characters**

Screenshot 2023-06-01 at 3 18 16 PM

jonwaldstein (approved):

jonwaldstein:
@kjohnson made 2 updates:

1. fixed the quill editor tooltip by adding a `bounds` prop [5078072](https://github.com/impress-org/givewp-next-gen/pull/181/commits/50780724d18a056e6566e5963aa026e721521c64)
2. fixed the output of tag descriptions by applying  `html_entity_decode` [3834616](https://github.com/impress-org/givewp-next-gen/pull/181/commits/383461635f9bf739068cf4fcacf3f5772b6694f1)

PR #182, opened 2023-05-11
Refactor: update WordPress compatibility to v5.9 and load styles from WordPress

jonwaldstein:
## Description

Since we moved to WP Scripts and are loading dependencies from WordPress, we need to update our compatibility to support specific versions of WP and Gutenberg.

**Notes**

- Apparently if you add a dependency to `wp_enqueue_script` that has not been registered - the page does not load?  I noticed a package called `wp-preferences` did not exist in one of the earlier versions but is not a show stopper for the form to load - so we're now looping through dependencies and checking if they're registered first before adding them.  If not we're logging it as _missing_ and removing from the array given to enqueue.  I'm sure we can get better at identifying missing dependencies and loading them externally somehow but what I have here works fine for now.
- Once I went below 5.8 - the `ShortcutProvider` was missing which seems like the main interface component was depending on.  If we _really_ need to go lower than 5.9, i'm sure we can figure out a work-around but it did not work out of the box.
- By default `<RichText` comes with formatting options like Bold, Italic, Links which will eventually be useful.  I'm disabling for now with `<RichText allowedFormats={[]} />` which works across different Gutenberg versions.

## Affects

- The form builder now loads scripts and styles from WordPress.  There was also a bunch of minor styling tweaks.
- Update WP requirement to 5.9

## Visuals

https://www.loom.com/share/2a4aa087073445f284308fe88e4a9b0a

## Testing Instructions

- Pull branch and run `npm install && npm run dev`
- Use WP CLI to switch between WP versions checking between 5.9 - 6.2 `wp core update --version=6.2 --force`
- Load a next gen form builder and make sure it looks okay.

## Pre-review Checklist

-   [ ] Includes unit tests


kjohnson (approved):
Looks good to me.

image

JasonTheAdams (approved):

PR #185, opened 2023-05-16
Feature: donation form block render filter

jonwaldstein:
## Description

🛑 This is a breaking change for block names, existing forms will need to be deleted and re-created.

This adds a new api for integrating custom fields/blocks into the donation form.

New hook `add_filter('givewp_donation_form_block_render_${blockNameWithNamespace}', $block, $blockIndex)`

Using an existing field type

```
use Give\Framework\FieldsAPI\Text;
use Give\NextGen\Framework\Blocks\BlockModel;

add_filter(
    'givewp_donation_form_block_render_givewp-fee-recovery/fee-amount',
    static function (BlockModel $block, int $blockIndex) {
        return Text::make('feeAmount');
    },
    10,
    2
);
```

Using a custom field type

```
use Give\Framework\FieldsAPI\Text;
use Give\NextGen\Framework\Blocks\BlockModel;

add_filter(
    'givewp_donation_form_block_render_givewp-fee-recovery/fee-amount',
    static function (BlockModel $block, int $blockIndex) {
        class FeeAmount extends Text {
            const TYPE = 'feeAmount';
        }

        return FeeAmount::make('feeAmount');
    },
    10,
    2
);
```

## Affects

- Renamed `custom-block-editor/` namespace to `givewp/` 🛑
- Renamed `custom-block-editor/field` to `givewp/text` 🛑 
- Renamed `custom-block-editor/donation-amount-levels` to `givewp/donation-amount` 🛑 
- Removed default text field to ensure custom blocks have full control
- Simplified block names without the id suffix

**Change highlight**
The default text field has been removed.  This decision was two-fold.
1. We don't want generic text fields to correspond to custom blocks
2. If an external block is de-activated, we don't want anything to render

## Visuals

Example of hard-coding a custom block in the default blocks.json schema and using the filter in a mu-plugin:

Screenshot 2023-05-18 at 3 15 10 PM

## Testing Instructions

Currently, the only way to test this is programmatically, by either temporarily adding a custom block to `blocks.json` or registering one, then use the filter to return a fields api Node/Field.

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein:
@pauloiankoski this is ready for review.

cc: @kjohnson

src/NextGen/DonationForm/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -78,58 +85,81 @@ protected function convertTopLevelBlockToSection(BlockModel $block, int $blockIn
...
      * @throws NameCollisionException
+     *
+     * @return Node|null
      */
-    protected function createNodeFromBlockWithUniqueAttributes(BlockModel $block, int $blockIndex): Node
+    protected function createNodeFromBlockWithUniqueAttributes(BlockModel $block, int $blockIndex)
     {
-        switch ($block->name) {
-            case "custom-block-editor/donation-amount-levels":
+        $blockName = str_replace("custom-block-editor/", '', $block->name);
```
pauloiankoski, line 113: @jonwaldstein My only concern on this is that we are ignoring namespaces. This can cause collisions in the future in case two distinct developers decide to create blocks with the same names. I'd keep the namespace here, either for the switch-case section or for the hook.

jonwaldstein: @pauloiankoski that's a great observation, honestly I was thinking every block would use the same namespace because I thought we _had_ to - but since you said that i'm wondering if external blocks can and _should_ use a different namespace to avoid conflicts.  This might be the time to change `custom-block-editor` to `givewp` and test out a different namespace for a custom block.

pauloiankoski: I totally agree with that.

jonwaldstein: @pauloiankoski this has been updated ea1afe73

The only thing i'm questioning is if we should continue with the `{block-name}-field` convention or omit the word `-field` all together.

Examples:

- `givewp/text-field` vs `givewp/text`
- `givewp/email-field` vs `givewp/email`
- `givewp/donation-amount-field` vs `givewp/donation-amount`
- `givewp-fee-recovery/fee-amount-field` vs `givewp-fee-recovery/fee-amount`

On one hand, there could be value in keeping primitive / core fields keeping the `-field` suffix to differentiate them.
On the other hand I could be convinced that it's redundant or not necessary with the use of namespaces.


I would like to agree on a naming convention before this breaking change.

cc: @kjohnson @JasonTheAdams

pauloiankoski: Gutenberg, for instance, has blocks simple: `core/cover`. I'd go without that redundancy and make `givewp/text` the standard. Waiting to hear from the others too tho.

jonwaldstein: That's a good point, the main difference for us is that some of our blocks are technically fields/inputs that interact with a form which is my only hesitation.  Still though, i'm not sure that really matters when naming a block under the hood.

jonwaldstein: @pauloiankoski i've decided to simplify the block names by omitting the suffix.  I would rather lean on namespaces and categories to identify the blocks than a naming convention for the block id itself.  Thanks for thinking through this with me.

jonwaldstein:
@pauloiankoski this is ready for re-review

PR #186, opened 2023-05-29
Feature: add Checkbox field template

JasonTheAdams:
## Description

This adds a Checkbox field to the donation form template system as a base field. There is no counterpart for the Form Builder as that will be added from FFM, this just makes it available programatically.

I also noticed that fields without any validation rules were being treated as a string and not allowing an "empty" value. Fields without validation just shouldn't be validated, so I fixed that.

## Visuals

image

## Pre-review Checklist


JasonTheAdams:
Here we go, @pauloiankoski! It's ready for review. If it looks good go ahead and merge it so you can use it in Fee Recovery.

pauloiankoski (approved):

PR #189, opened 2023-06-02
Fix: set the value properly for checkbox template

JasonTheAdams:
## Description

#186 added the checkbox template, but it didn't correctly add the necessary value input. I've since added this in https://github.com/impress-org/givewp/pull/6813.

As a note, the `checked` attribute _should not_ be applied since React Hook Form will set that based on whether the default value is equal to the input value.

## Affects

Only the checkbox template

## Pre-review Checklist


JasonTheAdams:
@jonwaldstein In light of the changes in https://github.com/impress-org/givewp/pull/6813 and the checkbox PR already being merged, this is necessary to properly set the value.

JasonTheAdams:
@pauloiankoski I believe you used the checkbox template in Fee Recovery? Does this change mean you'll need to change something over there?

pauloiankoski:
@JasonTheAdams I do use that component, but it's not part of the React Hook Form system. It's only used to enable the logic, not to pass any value to the backend. For that reason, I don't think we need to change anything on Fee Recovery.

CleanShot 2023-06-02 at 20 23 44

JasonTheAdams:
Interesting, @pauloiankoski. I would expect what you're doing there to throw TypeScript errors. You're not passing the `value` and the `inputProps` aren't complete compared to what it's expecting. I wouldn't change how you're using it, but I'd expect you to have to suppress some TS errors.

pauloiankoski:
@JasonTheAdams I'm probably not receiving any error because there are not types for the Checkbox component in the addon context.

JasonTheAdams:
Ah, I see. Got it. Well, it should still work.

JasonTheAdams:
@jonwaldstein I've resolved the merge conflict and merged the corresponding PR in core. This is ready for review!

src/DonationForms/resources/propTypes.ts
```diff
@@ -35,6 +35,10 @@ export interface RadioFieldProps extends FieldProps {
     options: Array<RadioOption>;
 }
 
+export interface CheckboxProps extends FieldProps {
+    value: string | number;
```
jonwaldstein, line 39: I think we should add boolean as a type, or will that be the default if value is not provided?

jonwaldstein: oh interesting, well I guess that makes sense.

jonwaldstein (approved):

PR #192, opened 2023-06-13
Refactor: add generic hook and additional parameter

JasonTheAdams:
## Description

This adds a generic `givewp_donation_form_block_render` hook and also adds a missing parameter to the existing hook. The filter was previously passing the block as the first parameter, but this isn't correct. For a filter the first parameter isn't context but the value being filtered. This makes it problematic for multiple functions firing on the same hook, and makes it impossible to chain hooks — as is being done now with the addition of the new hook.

Related: https://github.com/impress-org/give-fee-recovery/pull/323

## Affects

The `givewp_donation_form_block_render_{$blockName}` hook.

## Testing Instructions

I believe the only plugin using this hook so far is Fee Recovery. So make sure that still works with the new PR.

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
+++ b/src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
@@ -175,14 +175,23 @@ class ConvertDonationFormBlocksToFieldsApi
                             if (!is_user_logged_in()) {
                                 $field->required();
                             }
-                            
+
                             $field->rules(new AuthenticationRule());
                         }
                     });
 
             default:
                 $customField = apply_filters(
+                    'givewp_donation_form_block_render',
+                    null,
+                    $block,
+                    $blockIndex,
+                    $this->formId
+                );
+
+                $customField = apply_filters(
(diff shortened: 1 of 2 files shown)


src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -175,14 +175,23 @@ protected function createNodeFromBlockWithUniqueAttributes(BlockModel $block, in
...
                             }
-                            
+
                             $field->rules(new AuthenticationRule());
                         }
                     });
 
             default:
+                $customField = apply_filters(
+                    'givewp_donation_form_block_render',
```
jonwaldstein, line 185: @JasonTheAdams how are you anticipating using this generic filter vs the block name one? to return bulk fields or something?

JasonTheAdams: My preference is to always have a generic hook option in addition to specific hooks. My specific use-case was to be able to have a single hook function in FFM so all the block conversions can happen in a single file.

PR #194, opened 2023-06-13
Feature: add support to extend default blocks

pauloiankoski:
## Description

This pull request extends the functionality of default blocks.

It adds a new hook that runs when a new form is being created:
```php
Hooks::doAction('givewp_form_builder_new_form', $form);
```

And provides 4 new methods to the `BlockColletion` class to manage block injections:

```php
// Insert the $newBlock before the $refBlock at the specified $blockIndex.
insertBefore($refBlock, $newBlock, $blockIndex = 0);

// Insert the $newBlock after the $refBlock at the specified $blockIndex.
insertAfter($refBlock, $newBlock, $blockIndex = 0);

// Insert the $newBlock to the beginning of the array of inner blocks of $refBlock at the specified $blockIndex.
prepend($refBlock, $newBlock, $blockIndex = 0);

// Insert the $newBlock to the end of the array of inner blocks of $refBlock at the specified $blockIndex.
append($refBlock, $newBlock, $blockIndex = 0);
```
`$refBlock` can be either the name or clientId of the block to be used as a reference for positioning `$newBlock`.

## Testing Instructions
```php
// 1. Add an action to the new hook
add_action( 'givewp_form_builder_new_form', function($form) {

    // 2. Create the new block
    $block = BlockModel::make([
        'name' => 'givewp/paragraph',
        'attributes' => [
            'content' => 'Hello World!',
        ]
    ]);

    // 3. Use one of the methods above to position the new block
    $form->blocks->insertAfter('givewp/donation-summary', $block);
}); 
```

CleanShot 2023-06-13 at 19 13 28

## Pre-review Checklist

-   [x] Includes unit tests

### DIFF
--- a/src/Framework/Blocks/BlockCollection.php
+++ b/src/Framework/Blocks/BlockCollection.php
@@ -79,26 +79,27 @@ class BlockCollection implements Arrayable
     /**
      * @unreleased
      *
-     * @param string $nameOrId
-     * @param BlockCollection|null $blockCollection
-     * @param string $return self|parent
      * @return BlockCollection|null
      */
-    public function findBlock(string $nameOrId, BlockCollection $blockCollection = null, string $return = 'self')
+    public function findBlock(string $blockName, int $blockIndex = 0, BlockCollection $blockCollection = null, string $return = 'self', int &$count = 0)
     {
         if ($blockCollection === null) {
             $blockCollection = $this;
         }
 
         foreach ($blockCollection->blocks as $block) {
-            if ($block->name === $nameOrId || $block->clientId === $nameOrId) {
-                if ($return === 'self') {
-                    return $block->innerBlocks;
-                } elseif ($return === 'parent') {
-                    return $blockCollection;
+            if ($block->name === $blockName) {
+                $count++;
+
+                if ($count === $blockIndex + 1) {
+                    if ($return === 'self') {
+                        return $block->innerBlocks;
+                    } elseif ($return === 'parent') {
+                        return $blockCollection;
+                    }
                 }
             } elseif ($block->innerBlocks) {
-                $result = $this->findBlock($nameOrId, $block->innerBlocks, $return);
+                $result = $this->findBlock($blockName, $blockIndex, $block->innerBlocks, $return, $count);
                 if ($result) {
                     return $result;
                 }
@@ -111,16 +112,16 @@ class BlockCollection implements Arrayable
     /**
      * @unreleased
      */
-    public function insertBefore(string $blockNameOrId, BlockModel $block): BlockCollection
+    public function insertBefore(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
     {
-        $blockCollection = $this->findBlock($blockNameOrId, $this, 'parent');
+        $blockCollection = $this->findBlock($blockName, $blockIndex, $this, 'parent');
 
         if (!$blockCollection) {
             return $this;
         }
 
         $innerBlocks = $blockCollection->blocks;
-        $blockIndex = array_search($blockNameOrId, array_column($innerBlocks, wp_is_uuid($blockNameOrId) ? 'id' : 'name'));
+        $blockIndex = array_search($blockName, array_column($innerBlocks, wp_is_uuid($blockName) ? 'id' : 'name'));
         array_splice($innerBlocks, $blockIndex, 0, [$block]);
         $blockCollection->blocks = $innerBlocks;
 
@@ -130,16 +131,16 @@ class BlockCollection implements Arrayable
     /**
      * @unreleased
      */
-    public function insertAfter(string $blockNameOrId, BlockModel $block): BlockCollection
+    public function insertAfter(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
     {
-        $blockCollection = $this->findBlock($blockNameOrId, $this, 'parent');
+        $blockCollection = $this->findBlock($blockName, $blockIndex, $this, 'parent');
 
         if (!$blockCollection) {
             return $this;
         }
 
         $innerBlocks = $blockCollection->blocks;
-        $blockIndex = array_search($blockNameOrId, array_column($innerBlocks, wp_is_uuid($blockNameOrId) ? 'id' : 'name'));
+        $blockIndex = array_search($blockName, array_column($innerBlocks, wp_is_uuid($blockName) ? 'id' : 'name'));
         array_splice($innerBlocks, $blockIndex + 1, 0, [$block]);
         $blockCollection->blocks = $innerBlocks;
 
@@ -149,9 +150,9 @@ class BlockCollection implements Arrayable
     /**
      * @unreleased
      */
-    public function prepend(string $blockNameOrId, BlockModel $block): BlockCollection
+    public function prepend(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
     {
-        $blockCollection = $this->findBlock($blockNameOrId);
+        $blockCollection = $this->findBlock($blockName, $blockIndex);
 
         if (!$blockCollection) {
             return $this;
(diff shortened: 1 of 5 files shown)


src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,106 @@ public function getBlocks(): array
...
+                    return $block->innerBlocks;
+                } elseif ($return === 'parent') {
+                    return $blockCollection;
+                }
+            } elseif ($block->innerBlocks) {
+                $result = $this->findBlock($nameOrId, $block->innerBlocks, $return);
+                if ($result) {
+                    return $result;
+                }
             }
```
jonwaldstein, line 105: I think we need to think through this one a bit more.  The `clientId` property is something that changes often (every time the page loads I believe) so we don't want to use that for searching. The `name` property is not a unique attribute as it can be used across multiple blocks which means what we're actually doing here would be finding the first block with a matching name.

This method also looks like it's trying to compensate for being used internally and publicly which is adding complexity.

I would recommend scaling this back to block name to start and see how that simplifies this.  I would also add a method for finding by block index.

This is something that would also highly benefit from unit testing as your building the logic.

jonwaldstein (changes requested):

src/Framework/Blocks/BlockCollection.php
pauloiankoski: On 58e267727c469224d1e133ecc915d5885e3ee8bb, I removed the search by ID and implement the logic of $blockIndex so the block used as a reference will be the one presented for the "$blockIndex" time. For example:
```php
// The code below will add the $block as the last child of the third Section block.
$form->blocks->append('givewp/section', $block, 2);
```

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,107 @@ public function getBlocks(): array
     /**
      * @unreleased
      *
-     * @param BlockModel[] $blocks
-     * @param string $name
-     * @return BlockModel|null
+     * @return BlockCollection|null
      */
-    public function findByName(string $name, array $blocks = null)
+    public function findBlock(string $blockName, int $blockIndex = 0, BlockCollection $blockCollection = null, string $return = 'self', int &$count = 0)
```
jonwaldstein, line 84: Are all these extra params useful or meaningful as a public api of block collection? To me, everything except name & index seems very specific for the _other_ methods in the collection to work.  

If i'm using `findBlock` (or preferably `findByName`) its most likely just to get the block attributes.

It makes me think we should have a protected property like `findBlockRecursive()` (or `findByNameRecursive()` since we're searching by name)  that's used internally with all the extra logic and a public method `findByName()` that's more simple and could even use the recursive one under the hood if need be.

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,107 @@ public function getBlocks(): array
...
+    public function insertBefore(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
+    {
+        $blockCollection = $this->findBlock($blockName, $blockIndex, $this, 'parent');
+
+        if (!$blockCollection) {
+            return $this;
+        }
+
+        $innerBlocks = $blockCollection->blocks;
+        $blockIndex = array_search($blockName, array_column($innerBlocks, wp_is_uuid($blockName) ? 'id' : 'name'));
```
jonwaldstein, line 124: Is this leftover from the client id search?

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,107 @@ public function getBlocks(): array
...
+    public function insertAfter(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
+    {
+        $blockCollection = $this->findBlock($blockName, $blockIndex, $this, 'parent');
+
+        if (!$blockCollection) {
+            return $this;
+        }
+
+        $innerBlocks = $blockCollection->blocks;
+        $blockIndex = array_search($blockName, array_column($innerBlocks, wp_is_uuid($blockName) ? 'id' : 'name'));
```
jonwaldstein, line 143: Is this leftover from the client id search?

src/Framework/Blocks/BlockCollection.php
pauloiankoski: I've been thinking about that too. Will quickly change it.

pauloiankoski: Resolved b64be7ae8eaaaf93cca8f0898162979ccfc1705f

src/Framework/Blocks/BlockCollection.php
pauloiankoski: Resolved 1289cc03d34e67991132bfa4ba7beba5781ff219

src/Framework/Blocks/BlockCollection.php
pauloiankoski: Resolved 1289cc03d34e67991132bfa4ba7beba5781ff219

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,117 @@ public function getBlocks(): array
...
 
         return null;
     }
+
+    /**
+     * @unreleased
+     */
+    public function insertBefore(string $blockName, BlockModel $block, int $blockIndex = 0): BlockCollection
+    {
+        $blockCollection = $this->findByName($blockName, $blockIndex, 'parent');
```
jonwaldstein, line 127: I think we should use the internal find method for these methods so any changes to the public one will not affect these ones.

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,117 @@ public function getBlocks(): array
     /**
      * @unreleased
      *
-     * @param BlockModel[] $blocks
-     * @param string $name
-     * @return BlockModel|null
+     * @return BlockCollection|null
      */
-    public function findByName(string $name, array $blocks = null)
+    public function findByName(string $blockName, int $blockIndex = 0, string $return = 'self')
```
jonwaldstein, line 84: When do you anticipate needing to use this return param?

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,117 @@ public function getBlocks(): array
...
+            return $this;
+        }
+
+        $innerBlocks = $blockCollection->blocks;
+        $innerBlocks[] = $block;
+        $blockCollection->blocks = $innerBlocks;
+
+        return $this;
+    }
 }
```
jonwaldstein, line 195: Can you please update the `TestBlockCollection` unit test to include this new api?  I want us to be confident these work properly since they are going to be immediately used in add-ons.

src/Framework/Blocks/BlockModel.php
```diff
@@ -73,6 +73,18 @@ public function getAttributes(): array
         return $this->attributes;
     }
 
+    /**
+     * @unreleased
+     */
+    public function setAttribute(string $name, $value): BlockModel
```
jonwaldstein, line 79: Can you please update the `TestBlockModel` unit test to include this.

jonwaldstein (changes requested):
@pauloiankoski this is looking good, i'm just being extra cautious because we're going to be using this in add-ons.

src/Framework/Blocks/BlockCollection.php
```diff
@@ -79,24 +79,117 @@ public function getBlocks(): array
     /**
      * @unreleased
      *
-     * @param BlockModel[] $blocks
-     * @param string $name
-     * @return BlockModel|null
+     * @return BlockCollection|null
```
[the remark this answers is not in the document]
pauloiankoski: Actually: ```* @return BlockModel|BlockCollection|null```

src/Framework/Blocks/BlockCollection.php
pauloiankoski: Sure, no problem.

src/Framework/Blocks/BlockModel.php
pauloiankoski: Resolved baf7036aaa610288e16b4f37de0f017316ecbefe

src/Framework/Blocks/BlockCollection.php
jonwaldstein: since we're using the `findByNameRecursive` method for those now, do we still need that here for the public method?  I'm just trying to think of a situation where you would want to use `findByName` and return the parent collection vs just simply returning the block.

pauloiankoski: Gotcha!

pauloiankoski: Resolved 87d70a6e2644711c2028fca0e50bdc0c9aa86576

jonwaldstein (approved):
@pauloiankoski great job man! this is a great addition, thanks for working through all my suggestions 🙌

PR #199, opened 2023-06-15
Refactor: improve form builder styles

pauloiankoski:
## Description

This pull request resolves some styling issues in the form builder. The main changes are:
- Form Builder: reworked the positioning of content and sidebars on the screen so they flow as more sidebars are shown or the screen size is reduced, just like Gutenberg works. Previously, the content area was fixed and could sometimes be covered by a sidebar.
- Onboarding: centered the content in the first modal and replace the exit button variation in the steps modal from tertiary to link.
- General: removed font size definition from the body to ensure all line heights now use the default measurements.

## Visuals
Before
CleanShot 2023-06-15 at 20 08 43

After
CleanShot 2023-06-15 at 20 07 25

## Pre-review Checklist

-   [ ] Includes unit tests


jonwaldstein (changes requested):
@pauloiankoski thanks for these updates, I appreciate you attempting to improve the canvas and sidebar experience but i've gone down this path before and it comes with too many side effects that I don't want to introduce right now.

There's a balancing act that needs to happen on certain screen sizes that should prevent both sidebars from being open at the same time.  We've also talked about making the canvas  more fluid in terms of width.  The isolated block editor does a lot of this for us (if/when we make the switch)

For now, let's revert the changes to the canvas & sidebars so they remain fixed.  We'll circle back to this a little later once we agree on a direction

jonwaldstein (approved):
thank you! 🎨

PR #200, opened 2023-06-16
Feature: mount hooks to window for add-ons use

JasonTheAdams:
## Description

This PR mounts form builder hooks to the window so that add-ons can make use of them. The primary reason is so add-ons have a simple way of adding name fields and checking for conflicts.

An even better way of going about this, if not too difficult, would be to make use of `block.supports` with something like:
```js
{
    supports: {
        fieldNames: {
            generateFrom: 'label'
        }
    }
}
```

That would automatically add the name attribute and property to the inspector in the Advanced section. That would be awesome.

## Affects

Nothing, just new functionality.

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/App.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/App.tsx
@@ -6,6 +6,7 @@ import defaultBlocks from './blocks.json';
 import Feedback from '@givewp/form-builder/feedback';
 import {BlockInstance} from '@wordpress/blocks';
 import './App.scss';
+import mountHooksToWindow from '@givewp/form-builder/hooks/mountHooksToWindow';
 import FormBuilderErrorBoundary from '@givewp/form-builder/errors/FormBuilderErrorBounday';
 
 const {blocks: initialBlocks, formSettings: initialFormSettings} = Storage.load();
@@ -25,6 +26,8 @@ if (initialBlocks instanceof Error) {
 if (ShortcutProvider === undefined) {
     console.error('ShortcutProvider is undefined.');
 }
+
+mountHooksToWindow();
 
 export default function App() {
     return (
--- a/src/FormBuilder/resources/js/form-builder/src/hooks/mountHooksToWindow.js
+++ b/src/FormBuilder/resources/js/form-builder/src/hooks/mountHooksToWindow.js
@@ -1,0 +1,12 @@
+import {useFieldNameValidator} from './useFieldNameValidator';
+import {useSelectedBlocks} from './useSelectedBlocks';
+
+export default function mountHooksToWindow() {
+    window.givewp = window.givewp || {};
+    window.givewp.hooks = window.givewp.hooks || {};
+
+    window.givewp.hooks = {
+        useFieldNameValidator,
+        useSelectedBlocks,
+    };
+}


src/FormBuilder/resources/js/form-builder/src/supports/field-settings/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * @unreleased
+ */
+export default function registerHooks() {
+    addFilter('editor.BlockEdit', 'givewp/supports/field-settings-hoc', FieldSettingsHOC);
+    addFilter('blocks.registerBlockType', 'givewp/supports/field-settings-attributes', updateBlockTypes);
+
+    // Mounts the field settings inspector slots so other plugins can extend the inspector controls
+    window.givewp.slots = window.givewp.slots || {};
+    window.givewp.slots.FieldSettingsFill = FieldSettingsFill;
+    window.givewp.slots.DisplaySettingsFill = DisplaySettingsFill;
```
JasonTheAdams, line 19: I wasn't sure which TS Window instance to update to document these additions to the window.

JasonTheAdams:
@jonwaldstein I'd appreciate a review from you on this. It's a pretty significant addition to how new field blocks can be added more easily, and is necessary for the FFM work coming up.

JasonTheAdams:
@jonwaldstein This is ready for review again after adding support for defaults like we discussed.

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/slots.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @unreleased
+ */
+const {Slot: FieldSettingsSlot, Fill: FieldSettingsFill} = createSlotFill('GiveWP/FieldSettings/FieldSettingSlot');
+const {Slot: DisplaySettingsSlot, Fill: DisplaySettingsFill} = createSlotFill(
+    'GiveWP/FieldSettings/DisplaySettingSlot'
+);
+
+window.givewp.slots = window.givewp.slots || {};
+window.givewp.slots.FieldSettingsFill = FieldSettingsFill;
+window.givewp.slots.DisplaySettingsFill = DisplaySettingsFill;
```
jonwaldstein, line 16: I think we can remove this since your already assigning in registerHooks() right?

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/index.tsx
jonwaldstein: @JasonTheAdams actually can we change this to `window.givewp.form.slots`? That way its scoped to our form object along with `window.givewp.form.blocks`.  

Also, we are not isolated from the admin scripts so other admin defined objects like the legacy `Give` is defined on the window. I imagine we'll do something similar for the detail pages like `window.givewp.donation.data`, `window.givewp.donation.slots` so just want to make sure we scope our window objects to avoid any conflicts.

jonwaldstein (changes requested):
When I add a text field and try to publish changes without updating the label, I get this error:

Screenshot 2023-07-07 at 7 51 10 AM

It looks like the default field name `FIELD NAME` is missing when you first add a field.  Also "Save to Donor Record" is being enabled by default which should be false:

Screenshot 2023-07-07 at 7 55 13 AM

src/FormBuilder/resources/js/form-builder/src/blocks/fields/text/Edit.tsx
```diff
@@ -1,98 +1,23 @@
...
-        if (!isUnique) {
-            updateFieldName(suggestedName);
-        }
-    }, [fieldName, updateFieldName, validateFieldName]);
-
-    const enforceRequiredValue = useCallback(() => {
-        if (!fieldName) {
-            updateFieldName(label);
-        }
-    }, [fieldName, updateFieldName, label]);
```
jonwaldstein, line 34: It doesn't look like this was carried over properly

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/index.tsx
JasonTheAdams: Got it. Resolved in 3c3728a

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/slots.ts
JasonTheAdams: Oops! Good catch. Resolved in 460cbc8

src/FormBuilder/resources/js/form-builder/src/blocks/fields/text/Edit.tsx
JasonTheAdams: Wow! Strange that didn't throw an error! Good catch! Resolved in 154e07c

JasonTheAdams:
@jonwaldstein Great catch all around! I also made some pretty significant updates to how field names are enforced. It's a bit complex so I made a video: https://www.loom.com/share/0133270b26854122b8488b5d70aeacb3

@pauloiankoski I did update the shape of the support settings, which I'll update in the description. So you'll need to update this for the checkbox to work properly.

JasonTheAdams:
@pauloiankoski Oh, and the slots are now in window.givewp.**form**.slots

JasonTheAdams:
@jonwaldstein @jdghinson I just updated the "Field Name" field to "Meta Key" per the design, and got its behavior dialed in. I made a Loom video to make sure I've got it right. Please give it a watch: https://www.loom.com/share/df95d2bfcf234c4fb10ce2e3118b9bfe?sid=afdaa0b9-4f75-4ab0-a480-a3e3740bc7be

Thanks!

JasonTheAdams:
Hey @jonwaldstein!

I'd like to get this PR merged as there's quite a bit of FFM work dependent on it. Let me know if you have any questions!

pauloiankoski:
@JasonTheAdams @jonwaldstein I'm facing a conflict with one of the blocks on FFM. The Email block label defaults to "Email", so the generated meta key is `email`, which conflicts with the email field from core. Do you think that, beyond checking for naming conflicts, we should prefix at least some of the meta keys we don't have full control of, like in FFM, with `ffm_{label}` or `custom_meta_{label}`?

jonwaldstein:
@pauloiankoski great thoughts, we had a recent discussion on @JasonTheAdams's loom video about meta keys.  I do like the idea of using prefixes on first generation and I think we should also blacklist core ones.  Jason pointed out that our core meta keys are using an underscore prefix but i'm realizing now that is not the same thing as the field api name - which is where this conflict is actually coming from.  Honestly, it makes me wonder if the field name and meta key _should_ be separate.

Screenshot 2023-07-20 at 1 15 36 PM

JasonTheAdams:
@pauloiankoski @jonwaldstein This doesn't really tip me towards discreet meta vs name, yet, as I think that way still adds more complexity to think through.

I feel like the solution here would be to set the default meta key value to something like `custom_email` — keeping in mind this is only an issue if the user places the field and then _doesn't_ give it a unique label, which I'm guessing they will as to distinguish from the required email field. So, if I'm understanding this correctly, we just need to change the default meta key value for this field. Is that right?

As a note, I'd need to make a quick code change to make that possible for the meta key attribute specifically.

JasonTheAdams:
@jonwaldstein @pauloiankoski I just added a couple things (mostly) from our discussions:
- I switched all hyphens to underscores in how meta keys are generated and whats permissible
- Meta keys cannot start with an underscore
- There's now a static array of names the meta key cannot be
- A new custom field will now check its name upon insertion (this should also fix the duplicate issue)

One thing I did notice while working on this is the continued complexity of meta keys and field names. Namely, this spans to other generated nodes such as paragraphs. They don't have meta keys, but they do have names (in the Fields API). It actually won't be a problem at this time because element names use a hyphen convention (e.g. "pargraph-1"), and meta keys must use underscores.

Next, if an add-on adds a custom node, it won't be included in the list of names to protect against. I'm open to ideas. What's really tricky here is that we could run into an issue such as:
1. User adds custom field with "fee_recovered"
2. User installs Fee Recovery
3. Fee Recovery adds "Fee Recovery" block with said name
4. User adds block to form
5. Boom. 😕 

This suggests to me that we may need two systems:
1. A system that knows what field names are preserved, including blocks not even placed, yet
2. A system that checks the form when the editor is pulled up for any potential conflicts with existing custom fields and the blocks

I'm pretty sure that we're getting into outlier cases. What really worries me is that if a user does run into the issue, they'll be met with a confusing message upon saving. At the very least, I think we should catch conflict exceptions when saving and report back the message, so the user even has a fighting chance. Hahah!

jonwaldstein:
@JasonTheAdams good observations.  One simple solution for preventing conflicts with add-ons is we could have a naming convention for _our_ fields.  So instead of `FeeRecovery::make('feeRecovery')` we would do something like `_feeRecovery` if (underscores are reserved) or even reserve the word `givewp` and make it `givewpFeeRecovery`.  Note: this would only really work in the context of the form builder - not programmatically as we have a general naming conflict check that would throw an error.

(thinking out loud)
I want to make sure we're not digging ourselves into a hole by merging the concept of meta keys and field names.  Although, technically they can be used in the same way because they are both unique identifiers, it's possible that they are a separation of concerns.  Our concept of _name_ in the fields api, spans across fields, elements, and groups. It's a unique identifier that is stored on the node (aside: we've even talked about replacing the unique property with a node ID) because of nodes like elements.  In the context of our donation form, the field name gets serialized as the dom element ID and/or input name.  When a field node opts into `StoreAsMeta` it simply uses the field name as the meta key during persistance.  I appreciate how simple that is, but to play devil's advocate it's not very flexible and your stuck with whatever non-conflicting field name you were allowed.

src/FormBuilder/resources/js/form-builder/src/hooks/useFieldNameValidator.ts
```diff
@@ -0,0 +1,102 @@
...
+    const nextIncrement = increments.length ? Math.max(...increments) + 1 : 1;
+    return `${prefix}_${nextIncrement}`;
+};
+
+/**
+ * @since 0.1.0
+ */
+export const flattenBlocks = (block) => [block, ...block.innerBlocks.flatMap(flattenBlocks)];
+
+const builtInFieldNames = [
```
jonwaldstein, line 37: A couple more to add:
`subscription_id` and `additional_email` from donor and donation meta

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+            if (!isUnique) {
+                slugifiedName = suggestedName;
+            }
+
+            setAttributes({
+                fieldName: slugifiedName,
+                emailTag: attributes.storeAsDonorMeta
+                    ? `meta_donor_${slugifiedName}`
+                    : `meta_donation_${slugifiedName}`,
```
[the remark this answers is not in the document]
jonwaldstein: Also we need to make a note to check if these custom email tags work as expected or if we need to update our system to support them.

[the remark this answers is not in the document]
JasonTheAdams: Good catch! Resolved in 7289baee

Also, I added the email tag check to the QA board.

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
```diff
@@ -0,0 +1,264 @@
...
+export default FieldSettingsHOC;
+
+/**
+ * Renders the field settings inspector controls.
+ *
+ * @unreleased
+ */
+function FieldSettingsEdit({attributes, setAttributes, fieldSettings}) {
+    const validateFieldName = useFieldNameValidator();
+    const [fieldNameSet, setFieldNameSet] = useState(attributes.hasOwnProperty('fieldName'));
```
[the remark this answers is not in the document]
JasonTheAdams: Yeah, that's better. Resolved in 2b35ffb9

src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
```diff
@@ -0,0 +1,264 @@
...
+            updateFieldName();
+        } else {
+            updateFieldName(attributes.fieldName, true);
+        }
+    }, [attributes.fieldName, updateFieldName]);
+
+    useEffect(() => {
+        // The first time the field is rendered set the field name to make sure the default meta key doesn't conflict
+        // with any existing meta keys.
+        if (isNewField && !attributes.hasOwnProperty('fieldName')) {
```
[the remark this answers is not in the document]
JasonTheAdams: Good catch! Resolved in 2b35ffb9

src/FormBuilder/resources/js/form-builder/src/hooks/useFieldNameValidator.ts
JasonTheAdams: Thanks! Resolved in 715f8d5a

JasonTheAdams:
Thanks for the feedback, @jonwaldstein! This is ready for review again!

jonwaldstein (approved):
Great addition, working well!

src/FormBuilder/resources/js/form-builder/src/hooks/useFieldNameValidator.ts
```diff
@@ -0,0 +1,104 @@
...
+    'state',
+    'zip',
+    'login',
+    'donation-summary',
+    'donationType',
+    'subscriptionFrequency',
+    'subscriptionInstallments',
+    'subscriptionPeriod',
+    'subscription_id',
+    'additional_email',
```
jonwaldstein, line 61: @JasonTheAdams Actually one more to add `formId`

JasonTheAdams: Done. Resolved in 4a6ed741

PR #208, opened 2023-06-30
Feature: Migrate Forms

kjohnson:
## Description

This PR introduces a pipeline of migration steps for creating a v3 form from a v2 form and is intended to integrate with the Forms Table UI (updates for which are in a separate PR).

For development, testing, and troubleshooting the migration steps can be run from the command line and inspected one at a time.

`wp givewp form:migrate {$id} --step`

## Architecture

The form migration logic is organized into "steps" which are run sequentially via a "pipeline" and a step "processor". The pipeline organizes the steps in sequence and supports before/after hooks for debugging. Each step is conditionally handled by the step processor, which checks an optional `canHandle()` method of the step class.


alaca (approved):
I really like how you organized everything and that you are using the pipeline pattern. Looks very clean and solid.  Also, adding support for CLI is a great addition. Migrate endpoint supports only one ID per request, which is fine for now, but maybe it will be a good idea to add support for bulk actions now so that we don't have to change frontend requests later, but I'll leave that decision to you. 
The only thing that is missing in this PR are class comments with a short description of the class and of course, unreleased tags :smile:  
Other than that, this looks great. Excellent work @kjohnson

src/FormMigration/Controllers/MigrationController.php
```diff
@@ -0,0 +1,72 @@
...
+            ->process($payload)
+            ->finally(function($payload) {
+                $payload->formV3->save();
+                Log::info(esc_html__('Form migrated from v2 to v3.', 'give'), $this->debugContext);
+            });
+
+        return new WP_REST_Response(array('errors' => [], 'successes' => [
+            $payload->formV2->id
+        ]));
+//        return new WP_REST_Response([
```
alaca, line 63: We probably don't need this anymore.

alaca (approved):
Looks good :+1: 
The only missing things are class comments and unreleased tags. Also, reformat the code using GiveWP coding standards and we are good to :rocket:  
Great work @kjohnson

src/FormMigration/Controllers/MigrationController.php
kjohnson: Resolved, see https://github.com/impress-org/givewp-next-gen/pull/208/commits/d0ea6a15483db48004c227f6940659823955741e.

PR #211, opened 2023-07-07
Feature: add focus ring for blocks

glaubersilva:
## Description

This PR adds a focus ring for the blocks when they are selected. This feature is helpful because provides UI feedback about which block is selected.

## Affects

The form builder

## Visuals

image

image

## Testing Instructions

Click on any block in the form builder and you should see a blue focus ring indicating that the block is selected, check the attached screenshot for reference.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204927308757475


src/FormBuilder/resources/js/form-builder/src/styles/_block-editor.scss
```diff
@@ -116,6 +116,17 @@
...
 
+    .wp-block {
+
+        &.is-selected {
+            padding: 1.5rem;
+            border-style: solid;
+            border-width: thin;
+            border-radius: 0.25rem;
+            border-color: var(--wp-admin-theme-color);
+        }
```
jonwaldstein, line 127: My main hesitation with this is that it causes the blocks to become narrower when selected (due to the additional padding).  I think if we tap into the `::before` pseudo to add the ring without any movement this could work.  Here's something quick and dirty that I was thinking:


```suggestion
        &.is-selected::before {
            border-style: solid;
            border-width: thin;
            border-radius: 0.25rem;
            border-color: var(--wp-admin-theme-color);
            display: block;
            width: calc(100% + 1.5rem);
            content: '';
            position: absolute;
            height: calc(100% + 1.5rem);
            left: calc(-1.5rem / 2);
            top: calc(-1.5rem / 2);
        }
```

glaubersilva: I liked it, just change a bit to keep 1rem as "padding" on all sides.

image

glaubersilva:
@jonwaldstein Thanks for the explanations, now I better understand your concerns in regard to this. I did implement your suggestion with a slight difference, let me know if there is anything else.

jonwaldstein:
@glaubersilva thanks! looks good to me but i'm also not the designer.  I'm thinking since this will affect all block selections we should do a quick screen share with @jdghinson to get his thoughts.  I think in his earlier designs he had a selected ring but there was no padding - so we'll want to see how this interacts with different block types.

glaubersilva:
@jonwaldstein Thanks! I already create a new zip with these last changes and send it to @jdghinson via Slack for a re-review.

glaubersilva:
@jonwaldstein After @jdghinson reviewed it, we implemented a couple of cosmetic changes, but one specific required a bit of JS. So, I recorded a video to better explain why it was necessary, let me know your thoughts on it. Thanks! 

Video: https://www.loom.com/share/6031260278924155b4289a856a0417d1

src/FormBuilder/resources/js/form-builder/src/common/popover.js
```diff
@@ -0,0 +1,23 @@
...
+        }
+    });
+
+    window.addEventListener('click', (event) => {
+        const popover = document.querySelector('.components-popover.block-editor-block-popover');
+        if (!!popover) {
+            popover.style.top = '0rem';
+        }
+    });
+}
```
jonwaldstein, line 23: @glaubersilva haha this looks fun! 
Thanks for the video explaining why you were working on this.  Looking at this and what the post editor is doing, i'm pretty sure we're trying to compensate for css issue on the form builder.  The reason is the block popover is calculating it's position based on it's relation to the window/document and block - so the reason it would be off is up the chain in our document somewhere.

Doing a quick look I was able to see our `.interface-interface-skeleton` is not actually full height and has a position top - while our `interface-interface-skeleton__header` header container might not need to position fixed.

I guarantee fixing the document css / height issues will resolve the popover problem.

glaubersilva: @jonwaldstein Ohh thanks! I spent more time than necessary on this one. 🤦‍♂️

Now it's resolved in 7a679bf61554d286b3c1bc81eeb2540b2e921e67 and 791bb1bad4cb9cc69242ddc40602b1e4bb921a09

jonwaldstein: Haha awesome, thanks!

jonwaldstein (approved):
@glaubersilva great job man!

PR #216, opened 2023-07-11
Feature: add AdvancedSettingsSlot to HOC

pauloiankoski:
## Description

This pull request adds a new slot, `<AdvancedSettingsSlot>`, to the Field Settings HOC. Add-ons can use this slot to add their own custom settings to the _Advanced_ section in the Inspector.

Directly using the `<InspectorAdvancedConstrols>` slot causes a rendering conflict with the fields added by the HOC.

## Visuals
Example of usage:
CleanShot 2023-07-11 at 14 53 20

## Pre-review Checklist

-   [ ] Includes unit tests

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
+++ b/src/FormBuilder/resources/js/form-builder/src/supports/field-settings/FieldSettingsHOC.tsx
@@ -1,4 +1,4 @@
-import {InspectorControls, InspectorAdvancedControls} from '@wordpress/block-editor';
+import {InspectorAdvancedControls, InspectorControls} from '@wordpress/block-editor';
 import {useCallback, useMemo} from '@wordpress/element';
 import {getBlockSupport} from '@wordpress/blocks';
 import {slugify} from '@givewp/form-builder/common';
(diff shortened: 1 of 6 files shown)


JasonTheAdams:
> Directly using the `<InspectorAdvancedConstrols>` slot causes a rendering conflict with the fields added by the HOC.

Can you elaborate on what this means? What happens when you use this?

pauloiankoski:
@JasonTheAdams It's easier to show than describe:

Using the existing `<InspectorAdvancedControls>`:
CleanShot 2023-07-12 at 15 36 01

Using the proposed `<AdvancedSettingsFill>`:
CleanShot 2023-07-12 at 15 38 24

JasonTheAdams:
Gotcha. What's a bit confusing is that this slot would be inconsistent with the other slots, which append to the end of the section. Do we have a field in a design that's prior to the fields being added?

pauloiankoski:
CleanShot 2023-07-12 at 16 26 33

This is what we have in the designs, with variations to fewer fields but none with more than in the image above.

JasonTheAdams:
Ah, I see. Personally, I think the "default value" should be in the Field Settings, not the Advanced Settings, and only for _some_ fields. Fields such as dropdowns already have something like this taken into consideration.

pauloiankoski:
@JasonTheAdams It's in FFM area on Figma. Other than that, I oversee other issues we need to find a solution like the space for Options PanelBody and for the Conditional Logic PanelBody.

JasonTheAdams:
@pauloiankoski We don't need to do anything special for adding the Options and Conditional sections. Those can be added to the blocks as one would normally add inspector controls and the panel body.

JasonTheAdams:
@pauloiankoski We don't need to do anything special for adding the Options and Conditional sections. Those can be added to the blocks as one would normally add inspector controls and the panel body.

pauloiankoski:
@JasonTheAdams Added two new slots: `<AfterFieldSettingsSlot>` and `<AfterDisplaySettingsSlot>`. Also reimplemented `defaultValue` and `emailTag` as supported in the `fieldSettings`. Email tag will have much more logic being added to it later, but this can be a starting point.

CleanShot 2023-07-13 at 14 29 28

JasonTheAdams (approved):
Looks good! Thanks, @pauloiankoski!

PR #217, opened 2023-07-17
Feature: add "default level" option to the Donation amount and levels block

glaubersilva:
## Description

This PR adds a new "Default Level" property as a "Radio" setting on the side of each level input. Now is also possible to reorder the levels using the mouse to drag the items.

## Affects

The Donation amount and levels block

## Visuals

image

image

## Testing Instructions

1. On the DONATION OPTION, set the "Multi-level Donation" value;
2. Go to the "Donation Levels" section and you could see a "Radio" setting on the side of each level;
3. Test the new controls by choosing different default levels and removing or adding new ones. 

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204282271232272


src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -102,6 +103,17 @@ const Inspector = ({attributes, setAttributes}) => {
...
                 />
+                {priceOption === 'multi' && (
+                    <SelectControl
+                        label={__('Default Level', 'give')}
+                        options={levels.map((level, index) => ({
+                            label: formatCurrencyAmount(level),
+                            value: level,
+                        }))}
+                        value={defaultLevel}
+                        onChange={(defaultLevel) => setAttributes({defaultLevel})}
```
jonwaldstein, line 114: Since people have to the ability to update the donation levels list - I think we'll additionally need to programmatically update this value on donation level change to make sure the default value exists in the donation levels array.  Otherwise there's a chance this default level could be saved but not have a matching level.

jonwaldstein (changes requested):

JasonTheAdams:
The biggest thing I'm wondering here, is if something like this would be a better UI for defining the amounts and selecting a default, than adding another dropdown.

image

cc: @jdghinson

glaubersilva:
@JasonTheAdams What's the name of this component?

jdghinson:
> 

@JasonTheAdams The chosen option would have been effective if the default amount is visible in the form builder. However, with the radio in FFM, you can easily select it and see it correctly in the form builder. If we can make it visible like the picture below then we can go with the direction you suggested.


Screenshot 2023-07-20 at 4 29 21 PM

src/DonationForms/Actions/ConvertDonationAmountBlockToFieldsApi.php
```diff
@@ -64,7 +64,7 @@ public function __invoke(BlockModel $block, string $currency): DonationAmount
                 ->fixedAmountValue($block->getAttribute('setPrice'))
                 ->defaultValue(
                     $block->getAttribute('priceOption') === 'set' ?
-                        $block->getAttribute('setPrice') : 50
+                        $block->getAttribute('setPrice') : $block->getAttribute('defaultLevel')
```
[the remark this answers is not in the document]
glaubersilva: Resolved in fd2f386f31f9fb80c4d51a7256e690c80d5264ac

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/settings.tsx
```diff
@@ -18,6 +18,7 @@ const {
     setPrice,
     priceOption,
     levels,
+    defaultLevel,
```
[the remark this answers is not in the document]
glaubersilva: Resolved in 1a46a6e7d94a301c7da123b66f9f6036e9055166

glaubersilva:
@jonwaldstein Thanks for the review. I removed the selector, now we're using "Radio" settings on the side of each level to set the default one. I updated the PR description (and screenshots) with the new changes. Let me know if there is anything more.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -76,6 +77,30 @@ const Inspector = ({attributes, setAttributes}) => {
...
+
+    const setNewLevels = (levelsOptions: OptionProps[]) => {
+        const newLevels = levelsOptions.filter((option) => option.value).map((option) => option.value);
+        if (JSON.stringify(newLevels) !== JSON.stringify(levels)) {
+            setAttributes({levels: newLevels});
+        }
+    };
+
+    setNewLevels(levelsOptions); //Update levels when the reorder feature is used
+
```
jonwaldstein, line 103: @glaubersilva I'm seeing a few issues here. It looks like there are side effects happening that will be triggered during every render/re-render.  This _could_ be prevented by putting the functions being fired inside a `useEffect` hook. 

Also, from first glance, i'm a little confused why we have both 
`setLevelsOptions` and `setNewLevels`.  If levelOptions is just converting the raw levels to an option structure, I would elevate that into a simple conversion function instead of trying to store it in state.  We should just be left with `setNewLevels` and `setDefaultLevel` callbacks.  The drag and drop should trigger a prop in the Options component like `onChange` and/or `setOptions` and therefore call `setNewLevels`.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
```diff
@@ -139,58 +164,13 @@ const Inspector = ({attributes, setAttributes}) => {
...
-
-                            newLevels.push(nextLevel.toString());
-                            setAttributes({levels: newLevels});
+                    <Options
+                        currency={true}
+                        multiple={false}
+                        options={levelsOptions}
+                        setOptions={(options) => {
+                            setNewLevels(levelsOptions);
+                            setLevelsOptions(options);
```
jonwaldstein, line 173: Same idea from my previous comment, this should really just be updating the levels & the default level in attributes, not the actual options structure.  When passing to the options prop, that's the time you can convert the levels to options.

jonwaldstein (changes requested):

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
glaubersilva: Now I'm using the `useEffect` to prevent re-render and I also changed the names to avoid confusion. The `useEffect` also updates the options on the screen when it gets recorded. About the `useState`, we need that because the Option component requires that to be able to update the options on the screen.

src/FormBuilder/resources/js/form-builder/src/blocks/fields/amount/inspector/index.tsx
glaubersilva: In fact, it should only change the options structure because this component was designed to use a specific state for updating the options on the screen when the user adds a new option or reorder them.

glaubersilva:
@jonwaldstein Thanks for the review, I implemented the `useEffect` and left a couple of replies to your previous comments. Let me know if there is anything more.

JasonTheAdams (approved):
I love it, @glaubersilva! What an improvement!

PR #233, opened 2023-07-27
Refactor: Differentiate which migrated forms are also "redirected"

kjohnson:
## Description

This PR renames updates the form ID redirect helper function to differentiate between when a form is migrated and when that migrated form is to be redirected.

Since redirecting a form's shortcodes/blocks is an option of the "transfer" process, a migrated forms needs to marked as to whether or not the migrated form ID should also be "redirected".

## Affects

This assumes the addition of the `redirectedFormId` meta, which denotes that a migrated form is also to be "redirected".

## Visuals

image


alaca (approved):
:+1:

PR #236, opened 2023-07-28
Feature: add support for multipart/form-data and file field uploads

jonwaldstein:
Depends on https://github.com/impress-org/givewp/pull/6857, https://github.com/impress-org/givewp/pull/6859, https://github.com/impress-org/givewp/pull/6865

## Description

This replaces our main donation request to use `multipart/form-data` instead of `application/json` and handles storage of file fields.  This takes advantage of the web api [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) interface to build the data for the request.  Since we are using the native [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) api it will automatically set the header content-type to `multipart/form-data` when using FormData for the body.

**Additional context**
Typical text-based forms (like our v2 forms) use `application/x-www-form-urlencoded` for content-type.
> In human terms, this means: "This is form data that has been encoded into URL parameters."

In v3 forms we originally opted to send our data as `application/json` really out of convenience and parity between our client and server because react-hook-form passes our handleSubmit callback a json object of values.

File uploads are considered a [special case](https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data#a_special_case_sending_files) because they are binary data.  This means the other available content-types won't be able to send this type of data to the server.  This is where the `multipart/form-data` content-type comes in and handles sending our form data as binary data.  The only real disadvantage I can see here is adding extra bytes to the request and a little less flexibility for nested data values.

Now, we _could_ have created a separate endpoint to handle file uploads specifically and left the form request _as is_ to use json but that would require a lot of overhead for one field type which would have to bypass our main donation endpoints.  In fact, that's basically how form field manager works.  However, this solution requires no overhead with a simple change to the request content-type.  It works seamlessly with our framework and and can still run through our validation process and along with the main donation request.  It also brings back native PHP superglobals to retreive data through `$_POST` and `$_FILES` where as `application/json` requires a raw input stream using `file_get_contents( 'php://input' )`.

## Affects

Since we're updating the form request content-type this will affect v3 form submission.

## Visuals

N/A

## Testing Instructions

Make sure GiveWP is `>=` https://github.com/impress-org/givewp/pull/6857 & https://github.com/impress-org/givewp/pull/6859 & https://github.com/impress-org/givewp/pull/6865

- First make sure v3 form is being processed properly
- Then programmatically add a file field using the snippet below and make sure the file gets uploaded to the media library and the donation meta records the media ID.

```
use Give\Framework\FieldsAPI\DonationForm;

add_action('givewp_donation_form_schema', static function (DonationForm $form) {
    $file = \Give\Framework\FieldsAPI\File::make('custom_file')
           ->maxUploadSize(wp_max_upload_size())
           ->allowedMimeTypes(get_allowed_mime_types())
            ->label('Custom File');

    $form->insertAfter('donationAmount', $file);
});
```

## Pre-review Checklist

-   [ ] Includes unit tests


src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -21,20 +25,56 @@ public function __invoke(DonationForm $form, Donation $donation, array $customFi
...
                 } else {
-                    // save as donation meta
-                    give()->payment_meta->update_meta($donation->id, $field->getName(), $value);
+                    $value = $customFields[$fieldName];
+
+                    // TODO: it would be nice to have a way of serializing the value from the field api
+                    $value = apply_filters("givewp_store_custom_field_value_$fieldName", $value, $field, $donation);
+
+                    // TODO: it would be nice to have a way of customizing the storage from the field api
+                    do_action("givewp_store_custom_field_$fieldName", $value, $field, $donation);
```
jonwaldstein, line 46: I plan on removing this in favor of a native fields api approach but wanted to showcase where something like this would be useful.  We recently discussed updates to the fields api to support serializing values and/or providing scopes for storage.

src/Framework/Routes/Router.php
```diff
@@ -68,13 +69,21 @@ protected function getDataFromPostRequest(): array
...
+        // v2 forms use "application/x-www-form-urlencoded"
+        // v3 forms use "multipart/form-data" and/or "application/json"
+        if (str_contains($contentType, "application/x-www-form-urlencoded") || str_contains(
+                $contentType,
+                "multipart/form-data"
+            )) {
+            $requestData = array_merge(
+                give_clean($_REQUEST),
+                //TODO: check if this is necessary for validation to find the value from the request
+                give_clean($_FILES)
```
jonwaldstein, line 81: I also plan on adding a validation rule for file upload fields which this would be necessary for to retrieve the value.

jonwaldstein:
@JasonTheAdams can you please give this a preliminary look over? it's not completely finished and I added a few comments that point out what needs to be updated but want to get your eyes on it.  I tried to explain what's happening in the description the best I could but would be happy to jump on a call to discuss further.

src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -21,20 +25,56 @@ public function __invoke(DonationForm $form, Donation $donation, array $customFi
...
 
-                $value = $customFields[$field->getName()];
-
-                if ($field->shouldStoreAsDonorMeta()) {
-                    // save as donor meta
-                    give()->donor_meta->update_meta($donation->donorId, $field->getName(), $value);
+                /** @var File $field */
+                if (($fieldType === Types::FILE) && isset($_FILES[$fieldName])) {
+                    //TODO: let file field provide this storage logic
+                    $this->handleFileUpload($field, $donation);
```
JasonTheAdams, line 38: This is happening _after_ validation, right?

src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -21,20 +25,56 @@ public function __invoke(DonationForm $form, Donation $donation, array $customFi
...
                 } else {
-                    // save as donation meta
-                    give()->payment_meta->update_meta($donation->id, $field->getName(), $value);
+                    $value = $customFields[$fieldName];
+
+                    // TODO: it would be nice to have a way of serializing the value from the field api
+                    $value = apply_filters("givewp_store_custom_field_value_$fieldName", $value, $field, $donation);
```
JasonTheAdams, line 43: Serializing the value in what context?

src/DonationForms/resources/registrars/templates/fields/File.tsx
```diff
@@ -0,0 +1,25 @@
...
+            <Label />
+
+            <input
+                type="file"
+                aria-invalid={fieldError ? 'true' : 'false'}
+                placeholder={placeholder}
+                onChange={(e) => {
+                    setValue(name, e.target.files[0]);
+                }}
+            />
```
JasonTheAdams, line 18: I wouldn't be opposed to using something like [React Dropzone](https://react-dropzone.js.org/) as a nice, new UI.

src/DonationForms/Listeners/StoreCustomFields.php
```diff
@@ -21,20 +25,56 @@ public function __invoke(DonationForm $form, Donation $donation, array $customFi
...
+    {
+        if ($field->shouldStoreAsDonorMeta()) {
+            // save as donor meta
+            give()->donor_meta->update_meta($donation->donorId, $field->getName(), $value);
+        } else {
+            // save as donation meta
+            give()->payment_meta->update_meta($donation->id, $field->getName(), $value);
+        }
+    }
 }
```
JasonTheAdams, line 80: This will be a great place to introduce scopes in a subsequent PR.

src/DonationForms/Listeners/StoreCustomFields.php
JasonTheAdams: I know what you're referring to regarding scopes, but I'm not sure what you're referring to for support for serializing values.

JasonTheAdams (commented):
> The only real disadvantage I can see here is adding extra bytes to the request and a little less flexibility for nested data values.

Nooooooo

This direction looks good to me! I much prefer it to the double-request method of before. I like files fitting right into the Field API flow like anything else. 💪

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: yup! that's one more advantage to keeping everything in one flow because the fields will run through the same validation - we just need to add a file rule

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: The context is for storage. Suppose you are sending an array to the server but want to store it as a pipe delimited string.  Right now the only way to do that would be to use a custom validation rule and call `serialize()`.  This might be a non issue with a fields api storage scope thing but could also be useful if you just want to opt into saving as donation meta but modify the value before storing.

src/DonationForms/resources/registrars/templates/fields/File.tsx
jonwaldstein: Yeah we've used that in the past and it sure was nice.  I'd be cool with circling back on the actual file field UI if people would prefer that - or maybe it's actually a form field manager setting

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: 💯

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: I mean to say `sanitize` explained in [this](https://github.com/impress-org/givewp-next-gen/pull/236#discussion_r1278205373) comment

jonwaldstein:
@JasonTheAdams this is ready for review and can be used in conjunction with https://github.com/impress-org/givewp/pull/6859

cc: @pauloiankoski

src/DonationForms/Listeners/StoreCustomFields.php
JasonTheAdams: Validation rules are capable of sanitizing, is that sufficient? Or we're thinking more?

JasonTheAdams:
@jonwaldstein I believe this is holding up the FFM release. What do we need to wrap up here?

src/DonationForms/Listeners/StoreCustomFields.php
jonwaldstein: Yeah that's how we're getting around those now which is sufficient for the time being.  Can circle back to this discussion

jonwaldstein:
@JasonTheAdams this has been updated to use scopes & is ready for re-review.  Couple things I noticed - which I don't think needs to be resolved right now because everything is working properly but stuff to think through:

1. We don't have a way of adding a default scope to a field. As a workaround, I'm still having to check for file field type - which I honestly think is fine because that's the core file type.
2. After uploading a file, the final storage of a file attachment ID still relies on the `$field->shouldStoreAsDonorMeta()` to determine if it should store that in donor or donation meta.  Since `file` is now it's own scope - that would always default to donation meta.

jonwaldstein:
@JasonTheAdams okay per our discussion, I removed the check for file scope here 49ab5bcf.  File fields are now handled special - in order to upload their attachment, but they can still specify their scope for the final attachment ID value.

JasonTheAdams (approved):
Love it!

PR #239, opened 2023-07-30
Feature: add new fields to the Fields API and their templates

pauloiankoski:
Depends on https://github.com/impress-org/givewp-next-gen/pull/200 to be merged before this PR.

## Description

This PR adds new field templates that will be used initially by Form Field Manager:
- Date
- File Upload (WIP)
- MultiSelect
- Phone
- URL

## Affects

- Fields API
- Field templates

## Pre-review Checklist

-   [ ] Includes unit tests


JoshuaHungDinh:
Screen Shot 2023-08-07 at 3 56 21 PM
Screen Shot 2023-08-07 at 3 56 11 PM

Resolved: [7078651](https://github.com/impress-org/givewp-next-gen/pull/239/commits/7078651b7f88cfe3cc2455da9d68eff055d97019), [8587811](https://github.com/impress-org/givewp-next-gen/pull/239/commits/8587811b7b360864f0ca6c993814dc3e67f16c56), [251fcba](https://github.com/impress-org/givewp-next-gen/pull/239/commits/251fcbac00a33b2ff4782400c039d41fa3ed0410)

src/DonationForms/resources/registrars/templates/fields/File.tsx
```diff
@@ -0,0 +1,20 @@
+import {FieldHasDescriptionProps} from '@givewp/forms/propTypes';
+
+export default function File({Label, ErrorMessage, fieldError, description, inputProps}: FieldHasDescriptionProps) {
+    const FieldDescription = window.givewp.form.templates.layouts.fieldDescription;
+
+    return (
+        <label>
+            <Label />
+            {description && <FieldDescription description={description} />}
+            <input
```
kjohnson, line 10: Does this input need to be controlled by React Hook Form?

kjohnson: Or is it already registered (by means of the `...inputProps`)?

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
```diff
@@ -0,0 +1,21 @@
...
+import {PhoneProps} from '@givewp/forms/propTypes';
+
+export default function Phone({Label, ErrorMessage, fieldError, description, phoneFormat, inputProps}: PhoneProps) {
+    const FieldDescription = window.givewp.form.templates.layouts.fieldDescription;
+
+    return (
+        <label>
+            <Label />
+            {description && <FieldDescription description={description} />}
+            {phoneFormat === 'domestic' ? (
```
kjohnson, line 12: Are we only supporting input masking for this `domestic` format?
Should this formatting be informed by the user client locale?

src/DonationForms/resources/registrars/templates/fields/File.tsx
JoshuaHungDinh: Hey @kjohnson, I am actually slated to speak with Jon about the file upload tomorrow. He has been helping us get through how it should be handled. I can provide more details after I've spoken with him.

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
JoshuaHungDinh: Thats a great question. It would seem so. I was reading the NPM package details it looks like there is a known issue with browser's autofill that might be providing inconsistent behavior. Which may be the reasoning behind it. I actually did not get to speak to Paulo about this specific block.

kjohnson: Ok - I'm not too worried about this right now and we can revisit this later. I don't think it needs to hold up this PR.

src/DonationForms/resources/registrars/templates/fields/File.tsx
kjohnson: @JoshuaHungDinh following up on this. This looks like the last bit for this PR before we can get it merged.

src/DonationForms/resources/registrars/templates/fields/Phone.tsx
pauloiankoski: Hey @kjohnson and @JoshuaHungDinh, I'd like to add something to this discussion. I've been following a basic rule for all cases here: if something is not specified in the designs (such as date formats), we'll continue to use it as it works in v2.

jonwaldstein:
Call Summary:

@jonwaldstein and @JoshuaHungDinh 

Topic: Date Field

In v2 forms w/ form field manager, the date field has formatting options.  These formatting options correspond to what gets rendered in the date picker on the donation form.  Interestingly enough, the date selected by the donor w/ said formatting option gets saved to the database as such (the final formatted string).  The confirmation page then simply displays that string.  So if the format was `yyyy/dd/mm` the value stored in the database would be `2023/10/08`.

In v3 forms we are considering the same date format options for the block - but are currently storing the raw utc date string value (`"2023-08-18T07:00:00.000Z"`).  Since that's the case here, the output on the confirmation page is that raw date string.

We have a few decisions to make here for the future of date fields:

1. Should we even have the _formatting_ options? or should the format be derived by the donor's locale on the front-end and the admin's locale on the backend?  this is the direction we are moving with amounts as the formatting is vast across locales.
2. Should we be storing the formatted string to the database like v2 ffm?  it's definitely the simple option but probably the least flexible.  Although, date strings can also be used in date object factories - the time might not translate correctly. For context, WordPress posts have `post_date` and `post_date_gmt` stored in a consistent format of `'Y-m-d H:i:s'` which could also work.
3. Regardless of these decisions, the confirmation page _should_ have a concept of receipt line item field types in case the browser needs to format them appropriately. We can add the field type to each receipt line item and/or provide a filter for the value.

These decisions are split between product and engineering.  I'm writing this here for transparency but would encourage further discussion with our team.

☮️

src/DonationForms/resources/registrars/templates/fields/File.tsx
JoshuaHungDinh: Yea Jon had merged his updated branch with changes to include a hidden field. 21dc20fb02007c2ad04f29f5c0c40a50d581b5c0

kjohnson (approved):
@pauloiankoski thanks for the great work!
And @JoshuaHungDinh thanks for making the updates!

PR #241, opened 2023-08-01
Feature: Add missing block implementation

jonwaldstein:
## Description

This implements the missing block when a previously registered block is not available.

The way this works is that when a block is missing, we replace it with the missing blockType and store the original block type json from the database in the missing blockType's originalContent attribute.

Then, when the block is found again as a registered block, we replace the missing block with the stored originalContent attribute.

Note: this is currently pulling in the WordPress core missing blockType to achieve this.  After going through this, I think it might be better to supply our own version to prevent any conflicts in future WP versions.

## Affects

- The form builder and how it handles missing blocks

## Visuals

Screenshot 2023-08-01 at 2 58 49 PM

## Testing Instructions

- Install a block extension like fee recovery
- Create a v3 form, add the fee recovery block and save.
- Deactivate the fee recovery plugin and go back to the form
- You should now see a notice about the missing block
- Now try re-activating the plugin to see the original block
- Now go de-activate the plugin again and return to the form
- This time, make a change to the form and save
- Refresh the form builder and make sure that missing block notice is still there
- Once again, re-activate the plugin to make sure you see the original block in the form builder
- You should also see the field on the form front-end without needing to save changes.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1204747417903437

### DIFF
--- a/src/FormBuilder/resources/js/form-builder/src/common/registerBlocks.ts
+++ b/src/FormBuilder/resources/js/form-builder/src/common/registerBlocks.ts
@@ -10,13 +10,13 @@ import {__experimentalGetCoreBlocks} from '@wordpress/block-library';
  */
 const registerMissingBlock = () => {
     //TODO: should probably replace this with a custom block
-    const missingBlock = __experimentalGetCoreBlocks().filter((block) => {
+    const missingBlock = __experimentalGetCoreBlocks().find((block) => {
         return block.name === 'core/missing';
     });
 
-    if (missingBlock && missingBlock.length > 0) {
-        const missingBlockName = missingBlock[0].name;
-        missingBlock[0].init();
+    if (missingBlock) {
+        const {name: missingBlockName} = missingBlock;
+        missingBlock.init();
 
         setUnregisteredTypeHandlerName(missingBlockName);
     }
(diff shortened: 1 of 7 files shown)


jonwaldstein:
@kjohnson please review this when you get a chance.

src/FormBuilder/resources/js/form-builder/src/common/registerBlocks.ts
```diff
@@ -1,4 +1,26 @@
...
+
+/**
+ * Registers the missing block from WordPress core.
+ *
+ * @see https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/missing
+ *
+ * @unreleased
+ */
+const registerMissingBlock = () => {
+    //TODO: should probably replace this with a custom block
```
kjohnson, line 12: I agree with a custom block in this case. Since this will most likely be an add-on that was deactivated, we can handle the user messaging in a much more productive way.

src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
```diff
@@ -51,7 +52,9 @@ const HeaderContainer = ({
     const onSave = (formStatus: FormStatus) => {
         setSaving(formStatus);
 
-        dispatch(setFormSettings({formStatus}))
+        dispatch(setFormSettings({formStatus}));
+
+        revertMissingBlocks(blocks);
```
kjohnson, line 57: Ok, cool - this is how I thought it would work.

kjohnson (approved):
Looks good to me. @jonwaldstein thanks for digging into this. We can create an issue for replacing the `missing-block` with a custom block.

src/FormBuilder/resources/js/form-builder/src/containers/HeaderContainer.tsx
jonwaldstein: Pretty interesting right? As a note, my theory for having to do this ourselves is because we're using json and not the html delimited structure that WordPress uses.

PR #248, opened 2023-08-08
Feature: add donor comments block

jonwaldstein:
## Description

This adds the Donor Comments block to our core blocks.

## Affects

- This new block adds a textarea field that integrates with the core donate controller to update the comment property on the donation model
- A form migration step was included
- The block will also be added upon new form if enabled in global/default options

## Visuals

Screenshot 2023-08-09 at 3 28 29 PM

Screenshot 2023-08-09 at 3 28 36 PM

## Testing Instructions

- Add a donor comments block to a v3 form.
- Interact with the settings
- Go to the front-end and fill in the field
- Complete the donation and check the back-end to make sure it persists.
- You should also see a line on the confirmation page.

## Pre-review Checklist

-   [ ] Includes unit tests

---
  - https://app.asana.com/0/0/1205223433088674


jonwaldstein:
@glaubersilva this is ready for review.

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -131,6 +132,13 @@ protected function createNodeFromBlockWithUniqueAttributes(BlockModel $block, in
             case "givewp/donor-name":
                 return $this->createNodeFromDonorNameBlock($block);
 
+            //TODO: give_is_donor_comment_field_enabled() prevents this from showing up in admin
```
JasonTheAdams, line 135: So what are we supposed to do with this? What's the to-do exactly?

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
```diff
@@ -131,6 +132,13 @@ protected function createNodeFromBlockWithUniqueAttributes(BlockModel $block, in
...
+            case "givewp/donor-comments":
+                return Textarea::make('comment')
+                    ->label($block->getAttribute('label'))
+                    ->helpText($block->getAttribute('description'))
+                    ->rules('max:255');
```
JasonTheAdams, line 140: Is this an existing rule? Are comments currently restricted to 255 characters?

src/DonationForms/resources/propTypes.ts
```diff
@@ -40,6 +40,10 @@ export interface CheckboxProps extends FieldProps {
     value: string | number;
 }
 
+export interface TextareaProps extends FieldProps {
+    helpText?: string;
+}
```
JasonTheAdams, line 45: I don't think we should refer to this as "help text", it should just be a description, which I believe we already have an interface for.

JasonTheAdams (changes requested):
This was fast, @jonwaldstein! Just a couple small suggestions.

src/DonationForms/resources/propTypes.ts
JasonTheAdams: Oh, I see the issue. We don't have a "description" concern in the Fields API. Crud. And the Help Text is being used all over the place.

I'm inclined to say let's move this along here and revisit it after Next Gen is merged into core so we don't keep going back and forth. Sound good?

jonwaldstein: yeah helpText already exists, we can additionally add description when that's ready in core & just use that in the template.  I think it's fine for now though

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
jonwaldstein: I noticed we were setting this rule on a few text fields which I think is because of the mysql data type - although now that i'm looking at it - that's for the meta_key not the value.  this one is stored in meta which says is longtext so I don't think we would ever reach that limit.

Screenshot 2023-08-10 at 2 58 42 PM

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
jonwaldstein: it was just for a reminder.  for v2 forms you can enable/disable on the form which would open this gate. on v3 forms we only have the global enabled - so would likely need to hook into the form schema to see if the block exists and update the form meta.

jonwaldstein: Resolved 940cbf91

src/DonationForms/Actions/ConvertDonationFormBlocksToFieldsApi.php
JasonTheAdams: Right, longtext has a monstrously large limit as it's an external file. I'd put a limit of 5,000 or something as a sanity limit, but that's about it.

JasonTheAdams: We only need to be setting a 255 limit for columns that are varchar.

jonwaldstein: sounds good, resolved 6eddbce6

jonwaldstein:
@JasonTheAdams ready for re-review

JasonTheAdams (approved):
Great work, @jonwaldstein!

kjohnson:
Merge conflicts resolved.

